PR #8, opened 2015-09-08
Warning added to avoid silent unavailability of intellij tasks.

AmailP:
The first time that I have tried the plugin I was unable to use the intellij tasks because of the non standard location of my plugin.xml file. This warning would... warn the user of the otherwise silent unavailability of the intellij tasks.


zolotov:
Thanks!

PR #48, opened 2015-12-11
README: Fix alternativeIdePath default value

Turbo87:
This PR fixes the `alternativeIdePath` option default value rendering.


zolotov:
thanks

PR #49, opened 2015-12-11
Remove redundant plugin application in plugins usage

JakeWharton:


Turbo87:
@JakeWharton missed one: https://github.com/JetBrains/gradle-intellij-plugin/blob/c7a2a5cc81c17e7dd966077b96aa22be25efaf73/README.md#buildgradle :wink:

JakeWharton:
Oops! I'm out, will update later. Good spot.

zolotov:
@Turbo87 the second one is required, see https://plugins.gradle.org/plugin/org.jetbrains.intellij

zolotov:
@JakeWharton thanks

JakeWharton:
I would actually be supportive of eliminating all of the mechanism of adding the plugin from @Turbo87's linked code. That part of the README should only focus on usage, not how to apply the plugin.

Turbo87:
I think you mean a different "second" one than I do. The one at the bottom is definitely not needed.

PR #63, opened 2016-01-13
debugIdea task added

gmandnepr:
This pull request adds task to start idea with installed plugin in debug mode.


zolotov:
Why do you need this?

gmandnepr:
I was debugging idea to find what code is triggered by some of the actions.
Also it was useful to understand what data is stored in PSI tree. Sometimes it is easier to evaluate expression then trying to find it in PSI viewer.

zolotov:
And what're you going to use for debug?

gmandnepr:
Idea is used to debug other instance of idea started by the plugin.

zolotov:
Won't merged then. IDEA is able debug gradle task without hardcoding debug port and introducing new task type.

screenshot 2016-01-13 17 18 00

PR #78, opened 2016-02-19
Run instrumentation once all sources compiled

### LINKED ISSUE #77: Instrumentation fails on 'compileKotlin' task on mixed Kotlin+Java sources
I've [project](https://github.com/VladRassokhin/intellij-hcl/blob/master/build.gradle) with mixed Kotlin + Java sources.
Everything works fine in 0.0.39. In 0.0.40 and build from sources (ab9507d99) instrumentation fails with:
```text
:compileKotlin FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileKotlin'.
> @NotNull instrumentation failed for /Users/vlad/devel/idea-plugins/intellij-hcl/build/classes/main/org/intellij/plugins/hcl/terraform/config/codeinsight/ModelHelper.class: java.lang.RuntimeException: Operation 'visitMaxs' failed for org/intellij/plugins/hcl/terraform/config/codeinsight/ModelHelper.getConnectionProperties(): Class not found: org.intellij.plugins.hcl.psi.HCLProperty
  com.intellij.compiler.instrumentation.InstrumentationClassFinder$2: Class not found: org.intellij.plugins.hcl.psi.HCLProperty
```

[Full log (`gradle clean build --stacktrace --info`) see here](https://www.dropbox.com/s/llqv4y66ae85z8n/gradle.error.log?dl=0).

Note project have non-standard sources layout also local version builds against IDEA 14.1.5.

`org.intellij.plugins.hcl.psi.HCLProperty` is java class from 'gen' directory.

AFAIK, Kotlin compiles first, so `.class` files from Java sources would not exist at `compileKotlin` phase. Kotlin compiler itself generates and uses stubs from Java sources, so it's not problem. Seems javac2 instrumenter depends on `.class` files and cannot work with sources.

VladRassokhin:
Fixes #77

Quite ugly fix, since it relates on 'classes' and 'testClasses' tasks, but it works.

### DIFF
--- a/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
@@ -2,6 +2,7 @@ package org.jetbrains.intellij
 
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.Task
 import org.gradle.api.artifacts.Configuration
 import org.gradle.api.artifacts.ResolveException
 import org.gradle.api.file.ConfigurableFileTree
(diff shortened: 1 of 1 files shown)


VladRassokhin:
Maybe some additional (nullability) checks required after `findByName`

zolotov:
Merged manually. Thanks

PR #81, opened 2016-02-26
External plugins dependencies

gmandnepr:
## This pull request is not ready yet. 

Proposed extra configuration to add dependency on external plugin:
```
intellij {
    externalPlugins = [[
                           id: 'org.intellij.plugins.markdown',
                           version: '8.5.0.20160208',
                           type: 'zip',
                           unzipTarget: 'markdown'
                       ],
                       [....]]
}
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,7 @@ compileGroovy {
 repositories {
     maven { url 'http://repo.labs.intellij.net/central-proxy' }
     mavenCentral()
+    mavenLocal()
 }
 
 dependencies {
@@ -21,7 +22,7 @@ dependencies {
     compile 'org.jetbrains:annotations:13.0'
     compile 'org.codehaus.groovy.modules.http-builder:http-builder:0.7.1'
     compile 'org.apache.httpcomponents:httpmime:4.5.1'
-    testCompile 'org.spockframework:spock-core:1.0-groovy-2.3', {
+    testCompile 'org.spockframework:spock-core:1.0-groovy-2.4', {
         exclude module: 'groovy-all'
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,6 @@ dependencies {
     compile localGroovy()
     compile 'org.jetbrains:annotations:13.0'
 
-    compile 'org.codehaus.groovy.modules.http-builder:http-builder:0.7.1'
     compile 'org.jetbrains.intellij:plugin-repository-rest-client:0.3.16'
     compile 'org.jetbrains.intellij.plugins:intellij-plugin-structure:1.14'
     compile 'org.jetbrains.intellij.plugins:common-utils:1.18'
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,7 @@ dependencies {
     }
 }
 
-version = '0.0.42'
+version = '0.0.42-SNAPSHOT'
 group = 'org.jetbrains'
 
 pluginBundle {
--- a/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
@@ -130,6 +130,11 @@ class IntelliJPlugin implements Plugin<Project> {
         project.dependencies.add(JavaPlugin.RUNTIME_CONFIGURATION_NAME, [
                 group: 'com.jetbrains', name: IDEA_MODULE_NAME, version: version, configuration: 'runtime'
         ])
+
+        extension.externalPlugins.each {
+            //TODO use provided configuration
+            project.dependencies.add(JavaPlugin.COMPILE_CONFIGURATION_NAME, project.files(downloadExternalPlugin(project, it)))
+        }
     }
 
     private static void configurePatchPluginXmlTask(@NotNull Project project) {
@@ -290,5 +295,32 @@ class IntelliJPlugin implements Plugin<Project> {
         configurations.getByName(JavaPlugin.RUNTIME_CONFIGURATION_NAME).getAllDependencies().each closure
         configurations.getByName(JavaPlugin.COMPILE_CONFIGURATION_NAME).getAllDependencies().each closure
 
+    }
+
+    private static def downloadExternalPlugin(@NotNull Project project, @NotNull Map<String, String> plugin) {
+        def host = "http://plugins.jetbrains.com"
+
+        new File("$project.projectDir/externalPluginsLibs/").mkdirs()
+
+        new AntBuilder().get(src:"$host/plugin/download?pluginId=$plugin.id&version=$plugin.version", dest:"$project.projectDir/externalPluginsLibs/$plugin.id.$plugin.type", skipexisting:true)
+
+        return extractExternalPluginJars(project, plugin)
+    }
+
+    private static def extractExternalPluginJars(@NotNull Project project, @NotNull Map<String, String> plugin) {
+        if ("jar".equals(plugin.type)) {
+            return [new File("$project.projectDir/externalPluginsLibs/$plugin.id.$plugin.type")]
+        }
+
+        new AntBuilder().unzip(src: "$project.projectDir/externalPluginsLibs/$plugin.id.$plugin.type", dest: "$project.projectDir/externalPluginsLibs/", overwrite: "false")
+
+        def libsDir = new File("$project.projectDir/externalPluginsLibs/$plugin.unzipTarget/lib")
+
+        return libsDir.listFiles(new FilenameFilter() {
+            @Override
+            boolean accept(File dir, String name) {
+                return name.endsWith("jar");
+            }
+        });
     }
 }
--- a/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
@@ -212,7 +212,7 @@ class IntelliJPlugin implements Plugin<Project> {
             description = "Bundles the project as a distribution."
             group = GROUP_NAME
             baseName = extension.pluginName
-            from("$prepareSandboxTask.destinationDir")
+            from("$prepareSandboxTask.destinationDir/$extension.pluginName")
             into(extension.pluginName)
             dependsOn(prepareSandboxTask) 
         }
--- a/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
@@ -112,12 +112,12 @@ class IntelliJPlugin implements Plugin<Project> {
         def ivyFile = createIvyRepo(project, extension)
         def version = extension.version
 
-        project.configurations.create("provided")
+        def externalPluginScope = project.configurations.create("externalPluginScope")
 
         project.sourceSets.configure {
-            main.compileClasspath += project.configurations.getByName("provided")
-            test.compileClasspath += project.configurations.getByName("provided")
-            test.runtimeClasspath += project.configurations.getByName("provided")
+            main.compileClasspath += externalPluginScope
+            test.compileClasspath += externalPluginScope
+            test.runtimeClasspath += externalPluginScope
         }
 
         project.repositories.ivy { repo ->
@@ -141,7 +141,7 @@ class IntelliJPlugin implements Plugin<Project> {
         ])
 
         extension.externalPlugins.each {
-            project.dependencies.add("provided", project.files(downloadExternalPlugin(project, it)))
+            project.dependencies.add("externalPluginScope", project.files(downloadExternalPlugin(project, it)))
         }
     }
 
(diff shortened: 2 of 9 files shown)


build.gradle
```diff
@@ -21,12 +22,12 @@ dependencies {
...
     compile 'org.codehaus.groovy.modules.http-builder:http-builder:0.7.1'
     compile 'org.apache.httpcomponents:httpmime:4.5.1'
-    testCompile 'org.spockframework:spock-core:1.0-groovy-2.3', {
+    testCompile 'org.spockframework:spock-core:1.0-groovy-2.4', {
         exclude module: 'groovy-all'
     }
 }
 
-version = '0.0.42'
+version = '0.0.42-SNAPSHOT'
```
gmandnepr: This will be updated to 0.0.43 most likely

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
```diff
@@ -291,4 +304,31 @@ class IntelliJPlugin implements Plugin<Project> {
...
+
+        def libsDir = new File("$project.projectDir/externalPluginsLibs/$plugin.unzipTarget/lib")
+
+        return libsDir.listFiles(new FilenameFilter() {
+            @Override
+            boolean accept(File dir, String name) {
+                return name.endsWith("jar");
+            }
+        });
+    }
```
gmandnepr: This code downloads and unpacks plugin on each build. I really don't like this. Hope to find a way to install downloaded zips and jars to local ivy repository.

src/main/groovy/org/jetbrains/intellij/PrepareSandboxTask.groovy
```diff
@@ -24,12 +26,15 @@ class PrepareSandboxTask extends Sync {
...
+
         CopySpecInternal plugin = rootSpec.addChild()
-        classes = plugin.addChild().into("classes")
-        libraries = plugin.addChild().into("lib")
-        metaInf = plugin.addChild().into("META-INF")
+        classes = plugin.addChild().into("$extension.pluginName/classes")
+        libraries = plugin.addChild().into("$extension.pluginName/lib")
+        metaInf = plugin.addChild().into("$extension.pluginName/META-INF")
+
+        externalPlugins = rootSpec.into(Utils.pluginsDir(extension, inTests)).from("externalPluginsLibs").exclude("*.zip")
```
gmandnepr: At the moment all dependencies are packet into distribution zip. I'll fix that in next commits.

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
```diff
@@ -111,6 +112,14 @@ class IntelliJPlugin implements Plugin<Project> {
         def ivyFile = createIvyRepo(project, extension)
         def version = extension.version
 
+        project.configurations.create("provided")
```
zolotov: `provided` is wide-spread configuration name, it's better to set some more specific name, e.g. `intellij-plugins`

zolotov: also it makes sense to make this configuration invisible

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
```diff
@@ -111,6 +112,14 @@ class IntelliJPlugin implements Plugin<Project> {
...
+
+        project.sourceSets.configure {
+            main.compileClasspath += project.configurations.getByName("provided")
+            test.compileClasspath += project.configurations.getByName("provided")
+            test.runtimeClasspath += project.configurations.getByName("provided")
```
zolotov: Is this really needed in `test.runtimeClasspath`? If so, why it's not required for `main.runtimeClasspath`?

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
zolotov: `project.configurations.create("provided")` returns create configuration which is can be used later, e.g. while adding dependencies or expanding project classpath

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
```diff
@@ -291,4 +304,31 @@ class IntelliJPlugin implements Plugin<Project> {
         configurations.getByName(JavaPlugin.COMPILE_CONFIGURATION_NAME).getAllDependencies().each closure
 
     }
+
+    private static def downloadExternalPlugin(@NotNull Project project, @NotNull Map<String, String> plugin) {
+        def host = "http://plugins.jetbrains.com"
```
zolotov: Extract constant please

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
```diff
@@ -291,4 +304,31 @@ class IntelliJPlugin implements Plugin<Project> {
...
+
+        new File("$project.projectDir/externalPluginsLibs/").mkdirs()
+
+        new AntBuilder().get(src:"$host/plugin/download?pluginId=$plugin.id&version=$plugin.version", dest:"$project.projectDir/externalPluginsLibs/$plugin.id.$plugin.type", skipexisting:true)
```
zolotov: `project.ant` instead of `new AntBuilder`

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
```diff
@@ -291,4 +304,31 @@ class IntelliJPlugin implements Plugin<Project> {
...
+    }
+
+    private static def extractExternalPluginJars(@NotNull Project project, @NotNull Map<String, String> plugin) {
+        if ("jar".equals(plugin.type)) {
+            return [new File("$project.projectDir/externalPluginsLibs/$plugin.id.$plugin.type")]
+        }
+
+        new AntBuilder().unzip(src: "$project.projectDir/externalPluginsLibs/$plugin.id.$plugin.type", dest: "$project.projectDir/externalPluginsLibs/", overwrite: "false")
+
+        def libsDir = new File("$project.projectDir/externalPluginsLibs/$plugin.unzipTarget/lib")
```
zolotov: Shouldn't returns jar files from `lib` only, all Zip content is required

src/main/groovy/org/jetbrains/intellij/PrepareSandboxTask.groovy
```diff
@@ -24,12 +26,15 @@ class PrepareSandboxTask extends Sync {
         group = IntelliJPlugin.GROUP_NAME
         description = "Creates a folder containing the plugins to run Intellij IDEA with."
 
+        def extension = project.extensions.findByName(IntelliJPlugin.EXTENSION_NAME) as IntelliJPluginExtension
```
zolotov: task.project.extensions.getByType(IntelliJPluginExtension)

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
zolotov: It works badly if previous downloading was interrupted

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
```diff
@@ -291,4 +304,31 @@ class IntelliJPlugin implements Plugin<Project> {
         configurations.getByName(JavaPlugin.COMPILE_CONFIGURATION_NAME).getAllDependencies().each closure
 
     }
+
+    private static def downloadExternalPlugin(@NotNull Project project, @NotNull Map<String, String> plugin) {
+        def host = "http://plugins.jetbrains.com"
+
+        new File("$project.projectDir/externalPluginsLibs/").mkdirs()
```
zolotov: Should be system-level cache, not project-level

src/main/groovy/org/jetbrains/intellij/PrepareSandboxTask.groovy
```diff
@@ -118,7 +123,7 @@ class PrepareSandboxTask extends Sync {
     }
 
     private void configureClasses(@NotNull IntelliJPluginExtension extension, boolean inTests) {
-        destinationDir = new File(Utils.pluginsDir(extension, inTests), "$extension.pluginName")
+        destinationDir = new File(Utils.pluginsDir(extension, inTests))
```
zolotov: This means that all plugins will be wiped on `cleanPrepareSandbox` task. And this is not good because you'll have to reinstall all plugins from debug IDE instance on each recompile

zolotov:
Thanks for doing this!

At the moment the issue has following requirements
- should be tolerant to interrupted downloads
- should fail with sensible message if plugin are not available
- should cache plugin on system level (in .gradle directory)
- should handle `zip` plugins properly – unpack in `${plugin-name}` directory and copy directory to `sandbox/plugins`
- at least couple common scenarios should be covered with tests
- should has better plugin dependency notation
  - do not require version number
  - download the latest compatible plugin with current intellij version (`org.jetbrains.intellij.PatchPluginXmlTask#sinceUntilBuild`)

Also please check that downloaded jars are not included into the target distribution of plugin.

Local ivy repo for plugins is not required.

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
gmandnepr: If dependency is included in main.runtimeClasspath then plugin is copied in to plugin libs. We expect that in runtime it will be provided by idea. As for the test, I am working on proper way to add that dependency in to plugins-test directory. After that change test.runtimeClasspath will not be required anymore

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
gmandnepr: True, but this is going to be only classes dependencies for compilation. For the test and runIdea phase I am going to unzip entire plugin into dedicated directory.

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
zolotov: In tests it also will be provided by IDEA

zolotov:
I've introduced `ExternalPluginRepository`, please rebase PR and try to use it in your implementation. I can do it if you have no time for it.

gmandnepr:
I will take a look during next 2-3 days.

src/test/groovy/org/jetbrains/intellij/ExternalPluginRepositorySpec.groovy
```diff
@@ -24,12 +24,12 @@ class ExternalPluginRepositorySpec extends IntelliJPluginSpecBase {
...
+        assert plugin.artifact.name == 'markdown'
         assert collectFilePaths(plugin.jarFiles, repository.cacheDirectory.absolutePath) ==
-                ['/plugins.jetbrains.com/org.intellij.plugins.markdown-master-8.5.0.20160208/lib/markdown-javafx-preview.jar',
-                 '/plugins.jetbrains.com/org.intellij.plugins.markdown-master-8.5.0.20160208/lib/Loboevolution.jar',
-                 '/plugins.jetbrains.com/org.intellij.plugins.markdown-master-8.5.0.20160208/lib/intellij-markdown.jar',
-                 '/plugins.jetbrains.com/org.intellij.plugins.markdown-master-8.5.0.20160208/lib/markdown.jar'] as Set
+                ['/plugins.jetbrains.com/org.intellij.plugins.markdown-master-8.5.0.20160208/markdown/lib/markdown-javafx-preview.jar',
+                 '/plugins.jetbrains.com/org.intellij.plugins.markdown-master-8.5.0.20160208/markdown/lib/Loboevolution.jar',
+                 '/plugins.jetbrains.com/org.intellij.plugins.markdown-master-8.5.0.20160208/markdown/lib/intellij-markdown.jar',
+                 '/plugins.jetbrains.com/org.intellij.plugins.markdown-master-8.5.0.20160208/markdown/lib/markdown.jar'] as Set
```
gmandnepr: I believe original download name may be required in some cases (markdown plugin seems to be looking for its css files by path markown/lib/default.css) so I have changed layout of the repository little bit. It is /plugins.jetbrains.com/**pluginId**-**pluginVersion**/**unzipped_downloaded_artifact** now

build.gradle
```diff
@@ -14,6 +14,7 @@ repositories {
     maven { url 'http://repo.labs.intellij.net/central-proxy' }
     maven { url 'http://dl.bintray.com/jetbrains/intellij-plugin-service' }
     mavenCentral()
+    mavenLocal()
```
zolotov: why?

src/test/groovy/org/jetbrains/intellij/ExternalPluginRepositorySpec.groovy
zolotov: good catch, thanks

build.gradle
gmandnepr: I do not have strict opinion about this. Just reuse artifacts that I have locally.

zolotov: I bet you set this repo after you run `gradle build` for this project, so all dependency were downloaded into gradle repo and mavenLocal won't be used anyway.

gmandnepr: Yes, that is true..

zolotov:
outdated

PR #91, opened 2016-04-28
Add SQLDelight to list of examples

AlecStrong:


zolotov:
This is the nice one, thanks. Could you also set url to actual plugin project (https://github.com/square/sqldelight/tree/master/sqldelight-studio-plugin) and add info that it's plugin for android studio? I think it would be useful for users who're looking for an example of Android Studio plugin.

AlecStrong:
Done!

zolotov:
Thanks!

PR #95, opened 2016-05-05
External plugins

gmandnepr:


zolotov:
Merged, thanks. I want to change some things and will publish plugin after that

PR #96, opened 2016-05-05
Update README.md to add new plugin usage

fmatosqg:


zolotov:
Please mark somehow that's the plugin is for Android Studio, I think it would be useful for developers who want to make one for AS

fmatosqg:
Thanks for reminding me of that.

PR #100, opened 2016-05-24
Fix suppression of instrumentation of Kotlin classes

udalov:
Instead of annotations in `kotlin/jvm/internal/` there is now only `kotlin/Metadata` (https://github.com/JetBrains/kotlin/commit/59dab0a558328b9df4f5540da60bd0dcd4b10a35).


zolotov:
@udalov thanks!

udalov:
Cool, thanks for the quick merge. BTW this was originally encountered in https://youtrack.jetbrains.com/issue/KT-12458

PR #140, opened 2016-10-18
README.md updated

chashnikov:
description of 'updateSinceUtilBuild' option corrected to reflect the actual behavior


zolotov:
Should be fixed in https://github.com/JetBrains/gradle-intellij-plugin/blob/master/src/docs/configuration.md, sorry

zolotov:
applied manually

PR #148, opened 2016-11-12
Added option "intellij.addToolsJar" for adding tools.jar to dependencies

ingokegel:
The automatic dependency on tools.jar does not work if you use the project on multiple machines because it points to a path in the current Gradle JVM which is machine-specific. With this new option you can disable the automatic dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,9 @@ Empty value means that the IDE that was used for compiling will be used for runn
 distributions. If empty – Gradle cache directory will be used.
 **Default value**: `<empty>`
 
+- `intellij.addToolsJar` – whether to add tools.jar from the current JDK to the dependencies.
+**Default value**: `true`
+
 ### Patching plugin.xml
 
 *Available in SNAPSHOT only*
--- a/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
@@ -54,6 +54,7 @@ class IntelliJPlugin implements Plugin<Project> {
             sameSinceUntilBuild = false
             intellijRepo = DEFAULT_INTELLIJ_REPO
             downloadSources = true
+            addToolsJar = true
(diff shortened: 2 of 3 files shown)


zolotov:
> The automatic dependency on tools.jar does not work if you use the project on multiple machines because it points to a path in the current Gradle JVM which is machine-specific. 

What do you mean it doesn't work on multiple machine? `tools.jar` is part of compilation classpath, yes, it's part of JDK, which is also machinve-specific, so?

ingokegel:
In the gradle project that I have imported on Windows, I have this in the module file:

```
    <orderEntry type="module-library" scope="RUNTIME">
      <library>
        <CLASSES>
          <root url="jar://C:/Program Files/Java/jdk1.8.0_112/lib/tools.jar!/" />
        </CLASSES>
        <JAVADOC />
        <SOURCES />
      </library>
    </orderEntry>
```
On my Mac, I will have to re-import the project and the absolute path will change. Imported project files should be stable and not dependent on the OS.

zolotov:
1. `gradle-intellij-plugin` doesn't know anything about idea-project files, also it doesn't require IDEA as a IDE for development and shouldn't deal with its configuration.
2. There is no any reason to keep libraries-files and iml-files in VCS when you using Gradle project model. It's even not recommended: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

ingokegel:
>There is no any reason to keep libraries-files and iml-files in VCS when you using Gradle project model. 

You can have more complex projects where not all module files are from the gradle project. I have such a project. It's not so easy in that case.

How about adding a pathToToolsJar property then?

zolotov:
No, `pathToToolsJar` property doesn't make any sense, the `tools.jar` from SDK should always be used. This plugin has nothing to do with IDEA project configuration files and should not be changed in order to fix your project configuration problems in IDEA, Eclipse, Netbeans or whatever tool which can understand gradle project model

PR #151, opened 2016-11-17
Added change-notes patching

da1z:
Added patching chage-notes in plugin.xml

### DIFF
--- a/src/main/groovy/org/jetbrains/intellij/PatchPluginXmlTask.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/PatchPluginXmlTask.groovy
@@ -13,6 +13,7 @@ class PatchPluginXmlTask extends ConventionTask {
     private Object version
     private Object sinceBuild
     private Object untilBuild
+    private Object changeNotes
 
     @OutputDirectory
     File getDestinationDir() {
@@ -98,6 +99,16 @@ class PatchPluginXmlTask extends ConventionTask {
         this.untilBuild = untilBuild
     }
 
+    @Input
+    @Optional
+    String getChangeNotes() {
+        Utils.stringInput(releaseNotes)
+    }
+
+    void setChangeNotes(Object changeNotes) {
+        this.changeNotes = changeNotes
+    }
+
     @TaskAction
     void patchPluginXmlFiles() {
         def files = getPluginXmlFiles()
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/jetbrains/intellij/PatchPluginXmlTask.groovy
```diff
@@ -98,6 +99,16 @@ class PatchPluginXmlTask extends ConventionTask {
         this.untilBuild = untilBuild
     }
 
+    @Input
+    @Optional
+    String getChangeNotes() {
+        Utils.stringInput(releaseNotes)
```
zolotov: I think it should be `changeNotes` instead of `releaseNotes`

src/main/groovy/org/jetbrains/intellij/PatchPluginXmlTask.groovy
```diff
@@ -147,4 +159,15 @@ class PatchPluginXmlTask extends ConventionTask {
             }
         }
     }
+
+    static void patchChangeNotes(String changeNotes, Node pluginXml) {
```
zolotov: It's very similar to `patchDescription`, right? Maybe we can generalize them somehow?

src/main/groovy/org/jetbrains/intellij/PatchPluginXmlTask.groovy
```diff
@@ -98,6 +99,16 @@ class PatchPluginXmlTask extends ConventionTask {
...
 
+    @Input
+    @Optional
+    String getChangeNotes() {
+        Utils.stringInput(releaseNotes)
+    }
+
+    void setChangeNotes(Object changeNotes) {
+        this.changeNotes = changeNotes
+    }
```
zolotov: it's recommended to add one more setter like `void changeNotes(Object changeNotes)`. In this case end-user will be able to write something like

```
patchPluginXml {
  changeNotes "Something changed"
}
```

src/test/groovy/org/jetbrains/intellij/PatchPluginXmlSpec.groovy
```diff
@@ -33,6 +33,23 @@ class PatchPluginXmlSpec extends IntelliJPluginSpecBase {
...
 
+    def 'patch change notes'() {
+        given:
+        pluginXml << "<idea-plugin version=\"2\"></idea-plugin>"
+        buildFile << "version='0.42.123'\nintellij { version = '14.1.4' }\n"
+        buildFile << "patchPluginXml { changeNotes = 'change notes' }"
+        when:
+        def project = run(IntelliJPlugin.PATCH_PLUGIN_XML_TASK_NAME)
+        then:
+        outputPluginXml(project).text == """<idea-plugin version="2">
```
zolotov: It's a shame that tests don't work for you :( Anyway, the actual result is

```
<idea-plugin version="2">
   <change-notes>change notes</change-notes> 
   <version>0.42.123</version>
   <description>Plugin pluginDescription</description>
   <idea-version since-build="141.1532" until-build="141.*"/>
 </idea-plugin>
```

PR #158, opened 2016-12-14
Added ability to patch plugin id in plugin xml file

da1z:


zolotov:
Thanks! I will publish this in few hours

Issue #162, opened 2017-01-10
Possible to have runIdea open the surrounding project?


breandan:
I would like to use this plugin to generate a fully-configured IDE as the Gradle build artifact for a project. For example, when someone wants to contribute to a project, all they would need to do is run `git clone git@github.com:my-project folder-name; cd folder-name; ./gradlew runIdea -project .` and this would open the IDE with the surrounding project ready to go. Is this currently possible and if not, would you consider adding this as a feature?

breandan:
Looks like IntelliJ IDEA has some [command line options](https://www.jetbrains.com/help/idea/2016.3/opening-files-from-command-line.html). It would be nice to allow users to pass a project path in through the DSL to avoid navigating menu options after `./gradlew runIdea` - IDE just opens the project with the plugin loaded.

breandan:
I have a prototype of the `project <path>` DSL working locally, but the only issue is that IntelliJ IDEA does not detect JAVA_HOME when freshly installed without a prior sandbox. This requires the user to manually exit the initial Gradle import wizard and configure the global JDK. If the user needs to configure the IDE via the GUI, this somewhat defeats the purpose. Wondering if there is a way to supply this setting via a command line flag, or write it to the sandbox?

zolotov:
Won't fix, see explanation there in PR #164. 

The plugin is responsible for configuring IDEA SDK and running the IDEA, any other configuration of IDEA is on client's Gradle build. Fortunately. Gradle is very customizable.

PR #164, opened 2017-01-17
Adds projectDirectory to open a project

breandan:
As described in #162, I added a `projectDirectory` property under the `IntelliJPluginExtension` that accepts a directory to be passed as an argument to IntelliJ IDEA upon startup, opening a predefined project directory in the IDE. For example, it can be configured to open the surrounding project:

```groovy

apply plugin: 'org.jetbrains.intellij'

intellij {
  projectDirectory = project.projectDir.path
}
```

The intended use case is to immediately start writing code on a project, without installing, setting up and configuring the IDE. This feature is not only useful for plugin development, but for open source projects in general that want to establish a reproducible development environment for their contributors.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -89,8 +89,18 @@ task sourcesJar(type: Jar) {
     from sourceSets.main.allSource
 }
 
+task fatJar(type: Jar) {
+    baseName = project.name + '-all'
+    from (configurations.compile.collect { it.isDirectory() ? it : zipTree(it) }) {
+        exclude "META-INF/*.SF"
+        exclude "META-INF/*.DSA"
+        exclude "META-INF/*.RSA"
+    }
+    with jar
+}
+
 artifacts {
-    archives javadocJar, sourcesJar
+    archives javadocJar, sourcesJar, fatJar
 }
 
 uploadArchives {
--- a/build.gradle
+++ b/build.gradle
@@ -89,18 +89,8 @@ task sourcesJar(type: Jar) {
     from sourceSets.main.allSource
 }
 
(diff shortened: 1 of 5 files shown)


zolotov:
Since `runIdea` task is [JavaExec](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/JavaExec.html) task, you can pass any arguments there in the same manner:

```
runIdea {
  args 'path/to/project'
}
```

Since the suggestion is only about passing yet another optional argument to `runIdea` I don't see any value in extending `intellij`-expection dsl.

breandan:
@zolotov I still believe it would be a good idea to include `projectDirectory` as part of the DSL, as the command line parameter to open a project is [poorly documented](https://www.jetbrains.com/help/idea/2016.3/opening-files-from-command-line.html) (does not mention folders/projects), and is not well known. It would be convenient to expose this parameter to users of the `gradle-intellij-plugin`, or at least provide a written example to avoid having to look up through the documentation. Would you like me to submit this PR as a documentation example instead?

zolotov:
> as the command line parameter to open a project is poorly documented

This particular problem should be fixed in the documentation. 

> I don't see how an optional parameter pollutes the DSL

But it does, adding the parameter means adding more code, more documentation, supporting it and supporting something that I cannot control. `JavaExec` task is able to pass JVM parameters, but its API or documentation are not the right places for listing all possible JVM parameters for the sake of discoverability.

> Would you like me to submit this PR as a documentation example instead?

To `intellij-sdk` documentation – yes, please; to `gradle-intellij-plugin` – no.

PR #165, opened 2017-01-22
Update README.md

robohorse:


zolotov:
README.md is a generated file. Also, plugin configuration looks quite primitive, there are several plugins with similar configuration, so it won't be very useful as an example.

Won't merge, sorry.

PR #174, opened 2017-02-12
issue#173 runIdea set systemProperties

huoguangjin:
please check [issue#173 intellij.systemProperties does not work](https://github.com/JetBrains/gradle-intellij-plugin/issues/173)

thank you

### DIFF
--- a/src/main/groovy/org/jetbrains/intellij/tasks/RunIdeaTask.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/tasks/RunIdeaTask.groovy
@@ -127,6 +127,7 @@ class RunIdeaTask extends JavaExec {
     }
 
     def configureSystemProperties() {
+        systemProperties(getSystemProperties())
         systemProperties(Utils.getIdeaSystemProperties(getConfigDirectory(), getSystemDirectory(), getPluginsDirectory(), getRequiredPluginIds()))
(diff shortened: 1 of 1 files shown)


zolotov:
test is required

zolotov:
ok, looks like it's not easy to test `runIdea` task and it works fine for `test` task

huoguangjin:
I'm sorry to reply so late

I know little about groovy and spock, and keep learning these days
I try to write test for hours, and here are some ways to test `runIdea`:

1. add a test flag to `RunIdeaTask.groovy`, in `exec()` method, check the flag and skip calling `super.exec()`, at last check the setting of `runIdea`
2. mock(or hook) main class "com.intellij.idea.Main" and check the SystemProperties, is it possible?..
3. just add `runIdea.doLast { ... }` to `buildFile` and check the task info from `standardOutput`, this's not a good idea because there is no callback after idea set up..

Maybe the first one is the easiest, but it will make the source code a little bit unpretty😥...

If you have a better idea, please let me know
Thank you

zolotov:
Thanks for the update. Indeed, the first option is the easiest to implement and to maintain. You may want to implement it in another PR, but to be honest I'm fine without this test.

huoguangjin:
All right, thank you

Issue #171, opened 2017-02-09
Form class instrumentation is not run for Kotlin code


Minecrell:
Using gradle-intellij-plugin 0.2.0+ (I've tested 0.2.0 and 0.2.1), the form class instrumentation is not run for Kotlin code for me.

I've created a simple test project where I can reproduce the problem: https://github.com/Minecrell/intellij-test

It has a few Java, Groovy and Kotlin sources and a form component in the Kotlin source set: https://github.com/Minecrell/intellij-test/tree/master/src/main/kotlin/net/minecrell/test

To reproduce this problem I'm running `gradle clean runIdea`:

Using gradle-intellij-plugin 0.1.10 everything works correctly, however using 0.2.1 I get an error at runtime (which means that the form class instrumentation has not processed the Kotlin class):

```
Plugin 'net.minecrell.myplugin' failed to initialize and will be disabled.  Please restart IntelliJ IDEA.

kotlin.UninitializedPropertyAccessException: lateinit property panel has not been initialized
    at net.minecrell.test.TestComponent.<init>(TestComponent.kt:9)
```

In the Gradle log there is the following warning once for `:compileJava` and once for `:compileGroovy` (but not for `:compileKotlin`):

```
[ant:instrumentIdeaExtensions] /home/minecrell/dev/tmp/intellij-test/src/main/kotlin/net/minecrell/test/TestComponent.form: Class to bind does not exist: net.minecrell.test.TestComponent
```

Running gradle with the info flag suggests that it is not actually run for the Kotlin task: 

`Compiling forms and instrumenting code with nullability preconditions` appears only twice in the log, after `:compileJava` and `:compileGroovy` but not after `:compileKotlin`.

Minecrell:
I've taken a closer look at this. The root issue seems to be that `gradle-intellij-plugin` configures the instrumentation action for all `AbstractCompile` tasks the `classes` task depends on. However, it doesn't seem like the `classes` task does actually have a dependency on `compileKotlin`. Printing the dependencies reports the followiing:

```
[task ':compileJava', task ':processResources', task ':compileGroovy']
```

It works when I add a dependency on `compileKotlin` manually for the `classes` task:

```gradle
tasks.classes.dependsOn 'compileKotlin'
```

zolotov:
Hi @Minecrell,

Thank you for investigation. The fix will appear in `0.2.2` tomorrow. 

[...]

Issue #116, opened 2016-07-07
Form classes instrumentation always triggers Kotlin recompilation


matklad:
if I execute `./graldew assemble` twice, I expect the second time to be UP-TO-DATE. This unfortunately does not happen, and Kotlin is always recompiled. Running `./graldew assemble --info` gives the following hint:

```
:compileKotlin (Thread[Daemon worker Thread 2,5,main]) started.
:compileKotlin
Executing task ':compileKotlin' (up-to-date check took 0.035 secs) due to:
  Output file intellij-rust/build/classes/main/org/rust/ide/idea/CargoConfigurationWizardStep.class has changed.
  Output file intellij-rust/build/classes/main/org/rust/cargo/project/settings/ui/RustProjectSettingsPanel.class has changed.
  Output file intellij-rust/build/classes/main/org/rust/cargo/project/configurable/RustProjectConfigurable.class has changed.
All input files are considered out-of-date for incremental task ':compileKotlin'.
```

Those are Kotlin classes bound to the GUI designer forms.

zolotov:
Sorry, I doubt I can do anything with this

matklad:
Can't the instrumentation task be skipped if it should be up to date?

zolotov:
Instrumentation task is not task. It's kotlin task out-of-date

matklad:
Hm, maybe `doLast` hook should be used for instrumentation? 

I was able to workaround the issue by using `instrumentCode=false` and inspeat invoking the instrumentation manually in the `doLast`: https://gist.github.com/matklad/0685977255fad061d69bee8fa2a0e4dd

matklad:
Reference: https://discuss.gradle.org/t/how-do-i-configure-bytecode-instrumentation/5930/2

zolotov:
> Hm, maybe doLast hook should be used for instrumentation?

doLast is used for that

matklad:
>doLast is used for that

I mean that if I manually add instrumentation to the `doLast`, [like this](https://gist.github.com/matklad/0685977255fad061d69bee8fa2a0e4dd#file-uglyhack-gradle-L271-L275), then no recompilation happens and everything seem to work fine. If I just use `instrumentCode = true`, then the recompilations do happen. 

Perhaps intellij gradle plugin can provide a function which takes compile task as an argument, and instruments its output? That way the plugin user can add `doLast` to compile tasks manually and avoid spurious recompilation that way?

matklad:
[...]

Issue #73, opened 2016-02-03
Can't bind GUI designer form to a Kotlin class.


matklad:
Hi!

`IntelliJInstrumentCodeAction` uses only java `srcDirs`: https://github.com/JetBrains/gradle-intellij-plugin/blob/master/src/main/groovy/org/jetbrains/intellij/IntelliJInstrumentCodeAction.groovy#L30

Because of this it's impossible to bind a GUI designer form to a Kotlin class: the class is in `kotlin.srcDirs` and not in `java.srcDirs`.

I can work around this by using 
```
sourceSets {
    main.java.srcDirs += "src/main/kotlin"
}
```

PR #177, opened 2017-02-14
Move code instrumentation to extra task

Minecrell:
Up to now there were two different approaches in `gradle-intellij-plugin` to instrument the classes with form code and `@NotNull` assertions:

Previously, instrumentation was invoked using `classes.doLast`. However, since the files in the classes directory are changed after completion of the compile tasks, Gradle will assume they are out-of-date the next time it is run. Consequently, source files will be always recompiled. (see #116)

The current solution is to run instrumentation after each `AbstractCompile` task. While it fixes the up-to-date checks, there are several other problems with the current approach:

- It doesn't detect Kotlin properly (see #171)
- In some cases the class transformer requires other compiled classes, which might not be available when they're compiled by another compile task (see #73)
- Instrumentation runs separately for each compile task, but it always includes all classes. That means that classes are sometimes transformed several times:

  If I add https://github.com/Minecrell/intellij-test/commit/2164cd5ccc1a10a9e4a61b6e9a61a544375b92ea it transforms the class with `@NotNull` assertions twice, once with `:compileJava` and once with `:compileGroovy`: https://gist.github.com/Minecrell/0a1323e495280f96bf61e641d5b9010c (I'm not sure why it inserts two checks on each run)

  This is especially a problem when using Gradle's incubtaing incremental Java compilation: on each run new `@NotNull` assertions are added, even when the class was not modified since the last run.

### Conclusion
- Running instrumentation directly after each compile task isn't possible because it transforms some classes multiple times. It can also cause issues if required classes weren't compiled yet.
- Modifying the original classes after all compile tasks have run isn't possible because it breaks up-to-date checks.

### Solution
We need something that runs after all compile tasks, but without modifying the original classes.

I've added a new task that depends on `classes`. It runs after the `classes` task has finished, which means that all classes are already compiled but it will also not run when executing only the `classes` task or one of the compile task. It is, however, executed if a task depends on the source set output (e.g. the `jar` task) by adding it to `SourceSet.compiledBy`.

To keep the original classes unmodified it copies the original classes to a new directory (e.g. `build/classes/main-instrumented`). Then it transforms the classes using the Ant task.

Finally, all that is left is to make Gradle aware of the new classes directory. Changing `sourceSet.output.classesDir` works fine to include the instrumented classes in the JAR. As far as I could see in my testing, nothing breaks when changing the `classesDir`, except Kotlin's module check when compiling test classes. I've explained that in detail [in a comment in the source code](https://github.com/JetBrains/gradle-intellij-plugin/compare/master...Minecrell:extra-instrument-task?expand=1#diff-277af990756d301d731d16aca2d8d10bR245).

I'm not aware of any other problems when changing the `classesDir` at runtime, but I'm not entirely sure. All tests seem to pass when they are modified to run the correct tasks.


src/test/groovy/org/jetbrains/intellij/IntelliJPluginSpecBase.groovy
```diff
@@ -44,6 +44,12 @@ abstract class IntelliJPluginSpecBase extends Specification {
                 downloadSources = false
                 intellijRepo = '$intellijRepo'
             }
+
+            // Define tasks with a minmal set of tasks required to build a source set
+            sourceSets.all {
```
zolotov: I didn't get why this is needed

src/main/groovy/org/jetbrains/intellij/tasks/IntelliJInstrumentCodeTask.groovy
```diff
@@ -0,0 +1,108 @@
...
+import org.gradle.api.Nullable
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.file.FileCollection
+import org.gradle.api.file.SourceDirectorySet
+import org.gradle.api.tasks.*
+import org.jetbrains.annotations.NotNull
+import org.jetbrains.intellij.IntelliJPlugin
+import org.jetbrains.intellij.dependency.IdeaDependency
+
+class IntelliJInstrumentCodeTask extends DefaultTask {
```
zolotov: Could you make `git mv` please, it's hard to see the diff and it would be nice to keep some history

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
```diff
@@ -221,19 +222,48 @@ class IntelliJPlugin implements Plugin<Project> {
...
-        def abstractCompileDependencies = { String taskName ->
-            project.tasks.findByName(taskName).collect {
-                it.taskDependencies.getDependencies(it).findAll { it instanceof AbstractCompile }
-            }.flatten() as Collection<AbstractCompile>
-        }
-        abstractCompileDependencies(JavaPlugin.CLASSES_TASK_NAME).each {
-            it.inputs.property("intellijIdeaDependency", extension.ideaDependency.toString())
-            it.doLast(new IntelliJInstrumentCodeAction(false))
+        if (!extension.instrumentCode || extension.type == 'RS') {
+            return
```
zolotov: I'd added the task anyway and did nothing on `extension.instrumentCode || extension.type == 'RS'`

src/main/groovy/org/jetbrains/intellij/tasks/IntelliJInstrumentCodeTask.groovy
```diff
@@ -0,0 +1,108 @@
...
+    @SkipWhenEmpty
+    FileCollection getSourceDirs() {
+        def sourceDirs = existingDirs(sourceSet?.allSource)
+        sourceDirs?.removeAll(sourceSet.resources)
+        return project.files(sourceDirs)
+    }
+
+    @TaskAction
+    void instrumentClasses() {
+        // Clean output directory
```
zolotov: comment is redundant

src/main/groovy/org/jetbrains/intellij/tasks/IntelliJInstrumentCodeTask.groovy
```diff
@@ -0,0 +1,108 @@
...
+        output.deleteDir()
+
+        // Copy original classes
```
zolotov: extract to `copyOriginalClasses` method?

src/main/groovy/org/jetbrains/intellij/tasks/IntelliJInstrumentCodeTask.groovy
```diff
@@ -0,0 +1,108 @@
...
+                "$ideaDependency.classes/lib/asm-all.jar",
+                "$ideaDependency.classes/lib/jgoodies-forms.jar")
+
+        project.ant.taskdef(name: 'instrumentIdeaExtensions',
+                classpath: classpath.asPath,
+                loaderref: LOADER_REF,
+                classname: 'com.intellij.ant.InstrumentIdeaExtensions')
+
+        logger.info("Compiling forms and instrumenting code with nullability preconditions")
+        //noinspection GroovyAssignabilityCheck
```
zolotov: still needed?

src/main/groovy/org/jetbrains/intellij/tasks/IntelliJInstrumentCodeTask.groovy
```diff
@@ -0,0 +1,108 @@
...
+        boolean instrumentNotNull = prepareNotNullInstrumenting(classpath)
+
+        def srcDirs = sourceDirs
```
zolotov: inline

src/main/groovy/org/jetbrains/intellij/tasks/IntelliJInstrumentCodeTask.groovy
Minecrell: As far as I know `git mv` will only remove the file, rename it and add it again. Whether GitHub displays the diff depends on how similar the files are. I'm not sure if I can do much to make it display the file as diff.

zolotov: you right

> I'm not sure if I can do much to make it display the file as diff.

You can rename file and change its content in two different commits

src/test/groovy/org/jetbrains/intellij/IntelliJPluginSpecBase.groovy
Minecrell: In the previous version of the tests you've always executed only the minimal set of tasks that was necessary to do the tests. There is no task that builds a source set so I've added one. We could also execute `jar` or `assemble` but that would also build the JAR file and potentially run some other tasks. I can change it if you want.

Minecrell:
@zolotov I've added an extra rename commit. You can see the changes in the file in https://github.com/JetBrains/gradle-intellij-plugin/pull/177/commits/94f150f8805038e160beb483e4c47705d5de22a8#diff-18b17223bb923d44b86de5aec3cb7388L1 now

Minecrell:
I've pushed an additional commit that moves the task configuration out of `afterEvaluate` since that is no longer necessary. Let me know if I should squash it.

Minecrell:
I've just noticed there is a bug that will cause the wrong classes to be included when the task is up-to-date. The problem is that we modify the source set's `classesDir` in the `doLast` of the task, but that won't run when the task is up-to-date.

However, we can only change the `classesDir` after the task has already run so changing it on creation of the task isn't possible either. The only way I can think of right now is to add another task that runs always after the `instrumentCode` task and changes the source set. Can you think of a better way?

Minecrell:
I've pushed the additional task as described in my previous comment.

zolotov:
merged. thank you

Minecrell:
@zolotov Awesome, thank you! Let me know when you release a new a new version using these changes so we can update the plugin on our project.

zolotov:
@Minecrell 0.2.5 is release

Jonatha1983:
Hi @Minecrell 

About:

> I've just noticed there is a bug that will cause the wrong classes to be included when the task is up-to-date. The problem is that we modify the source set's `classesDir` in the `doLast` of the task, but that won't run when the task is up-to-date.
> 
> However, we can only change the `classesDir` after the task has already run so changing it on creation of the task isn't possible either. The only way I can think of right now is to add another task that runs always after the `instrumentCode` task and changes the source set. Can you think of a better way?

I have an idea on how to prevent the task out of date and wanted to know your opinion:  

Instead of the instrument task the plugin will add a doLast (if set to) to the compileJava task. The doLast will copy the original to some other folder - the instrument job will be done on the original folder - and in this way the task will benefit from gradle up-to-date. 

 You think it can work ?

PR #185, opened 2017-03-07
Prefer compileOnly for idea deps when it exists

### LINKED ISSUE #168: Plugin should use compileOnly instead of compile configuration for IDEA dependency
The current IDEA dependency leaks transitively from the project so it's hard to create a multi-module project where multiple modules depend on IDEA.

Also a lot of tools for shading / obfuscating include everything on the compile config default, pulling in the full IDEA dependency when doing so.

Gradle has a facility for this exact case - the `compileOnly` configuration (emphasis mine):

> Compile-only dependencies address a number of use cases, including:
> * Dependencies required at compile time but never required at runtime, such as source-only annotations or annotation processors;
> * Dependencies required at compile time but required at runtime only when using certain features, a.k.a. optional dependencies;
> * __Dependencies whose API is required at compile time but whose implementation is to be provided by a consuming library, application or runtime environment.__

(from https://blog.gradle.org/introducing-compile-only-dependencies)

madisp:
Fall back to compile configuration on older Gradle versions

Fixes #168

### DIFF
--- a/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
@@ -82,13 +82,25 @@ class IntelliJPlugin implements Plugin<Project> {
                 configureProjectAfterEvaluate(subproject, subprojectExtension)
             }
         }
-        configureIntellijDependency(project, extension)
-        configurePluginDependencies(project, extension)
+
+        Configuration idea = project.configurations.create('idea')
+        Configuration ideaPlugins = project.configurations.create('ideaPlugins')
+
+        if (JavaPlugin.hasProperty('COMPILE_ONLY_CONFIGURATION_NAME')) {
+            project.configurations.getByName(JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME).extendsFrom ideaPlugins
+            project.configurations.getByName(JavaPlugin.TEST_COMPILE_ONLY_CONFIGURATION_NAME).extendsFrom ideaPlugins
+        } else {
+            project.configurations.getByName(JavaPlugin.COMPILE_CONFIGURATION_NAME).extendsFrom ideaPlugins
+        }
+
+        configureIntellijDependency(project, extension, idea)
+        configurePluginDependencies(project, extension, ideaPlugins)
         configureTestTasks(project, extension)
     }
 
     private static void configureIntellijDependency(@NotNull Project project,
-                                                    @NotNull IntelliJPluginExtension extension) {
+                                                    @NotNull IntelliJPluginExtension extension,
+                                                    @NotNull Configuration configuration) {
         LOG.info("Configuring IntelliJ IDEA dependency")
         def resolver = new IdeaDependencyManager(extension.intellijRepo ?: DEFAULT_INTELLIJ_REPO)
         def ideaDependency
@@ -104,7 +116,7 @@ class IntelliJPlugin implements Plugin<Project> {
         }
         extension.ideaDependency = ideaDependency
         LOG.info("IntelliJ IDEA ${ideaDependency.buildNumber} is used for building")
-        resolver.register(project, ideaDependency)
+        resolver.register(project, ideaDependency, configuration)
 
         def toolsJar = Jvm.current().toolsJar
         if (toolsJar) {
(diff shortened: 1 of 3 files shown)


zolotov:
The same is needed for plugins as well

madisp:
cool, I'll update the patch and double-check the `runtime` configuration too

zolotov:
Thanks, looking forward to

Minecrell:
It might be better to just put them into a new configuration (e.g. `idea` or `intellij`, possibly a separate one for plugins?) and to extend that configuration in `compile`/`compileOnly`/`testCompileOnly`. That means you need to add the dependency a single time only once because it will be automatically inherited into the other configurations.

Basically like this (in pseudo Groovy):

```gradle
def ideaConfig = configurations.create('idea')
project.dependencies.add(ideaConfig.name, /* ... */)

if (compileOnly) {
    configurations.compileOnly.extendsFrom(ideaConfig)
    configurations.testCompileOnly.extendsFrom(ideaConfig)
} else {
    configurations.compile.extendsFrom(ideaConfig)
}
```

madisp:
oh true, that would make it a bit cleaner compared to shuttling the `List<String>` around

Minecrell:
Looks good, although after I thought about it for a while I think it might be useful in some cases to have a separate configuration for the IntelliJ IDEA plugins (e.g. `ideaPlugins`). For example if you have a task that generates PSI classes/parsers using [Grammar-Kit](https://github.com/JetBrains/Grammar-Kit) then you need just IntelliJ IDEA on the classpath when running it, not all the plugins you depend on.

All we need to do to support this is to create a new `ideaPlugins` configuration and extend the `idea` configuration in it. If you then extend `ideaPlugins` in `compileOnly` etc then it will automatically also inherit the `idea` configuration.

madisp:
so something like a `compileOnly > ideaPlugins > idea`? I'm not sure I understand the use case but I can send a new PR

Minecrell:
Yes. Essentially it allows re-using the IDEA dependency in other configurations. Grammar-Kit for example requires classes from the IntelliJ IDEA distribution for generating the parser/PSI classes. So if we have a separate `idea` configuration with only the `idea` dependency we can just do:

```gradle
configurations {
    grammarKit {
        extendsFrom configurations.idea
    }
}

dependencies {
    grammarKit("org.jetbrains:grammar-kit:...")
}

task generateParser(type: JavaExec) {
    classpath = configurations.grammarKit
    // ...
}
```

Currently everything is added to the `compile` configuration directly, so there is no easy way to only add the main IDEA distribution to the classpath of the Grammar-Kit task. With your changes in your PR here, we can already limit it to the IDEA dependency and the plugins you depend on. If you add another extra configuration for the IDEA plugins we can add only the required dependency on the IDEA core.

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
```diff
@@ -82,13 +82,25 @@ class IntelliJPlugin implements Plugin<Project> {
...
             }
         }
-        configureIntellijDependency(project, extension)
-        configurePluginDependencies(project, extension)
+
+        Configuration idea = project.configurations.create('idea')
+        Configuration ideaPlugins = project.configurations.create('ideaPlugins')
+
+        if (JavaPlugin.hasProperty('COMPILE_ONLY_CONFIGURATION_NAME')) {
+            project.configurations.getByName(JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME).extendsFrom ideaPlugins
```
Minecrell: Currently, you only extend `ideaPlugins`. You need to extend both `idea` and `ideaPlugins` or extend `idea` in `ideaPlugins`. I'm not entirely sure if it makes more sense to apply both separately or to extend `idea` in `ideaPlugins`...

To extend both you could do something like the following so you don't need to specify both for all 3 calls:

```groovy
Configuration[] ideaConfigs = [idea, ideaPlugins]

project.configurations.getByName(/*...*/).extendsFrom ideaConfigs
```

madisp: oh, my bad actually, I wanted to make ideaPlugins extend from idea. But I agree, I'm not sure if it makes sense, maybe better to keep the two configs separate?

Minecrell: I agree, I think it's better to keep them separate.

zolotov:
Thank you @madisp for the changes and @Minecrell for review. Merged. `0.2.6` will be available in a couple hours

Minecrell:
@zolotov Did you make the change I mentioned above? If not, only `ideaPlugins` is actually added to the classpath currently.

zolotov:
@Minecrell I did, thanks. Sorry, for delay, `0.2.6` is available at plugins.gradle.com now

madisp:
Thanks for the review and merge!

PR #189, opened 2017-03-28
Create configurations when applying the plugin

Minecrell:
When updating a project I noticed that the dependency configurations added in #185 are created in `afterEvaluate` instead of directly when the plugin is applied. While this is correct for adding the dependency, delaying the creation of the configuration means that it cannot be directly referenced in the build script (only in another `afterEvaluate` block).

The following build script will throw an error using `0.2.6`:

```gradle
task generateParser(type: JavaExec) {
    classpath = configurations.idea
    // ...
}
```


zolotov:
Thanks!

PR #190, opened 2017-03-29
Fix detecting compileOnly configuration

madisp:


zolotov:
btw it could be easy to test. Please see https://github.com/JetBrains/gradle-intellij-plugin/blob/master/src/test/groovy/org/jetbrains/intellij/IntelliJPluginSpecBase.groovy#L72.

Looks like it's easy to substitute gradle version in a particular test, create a task that will be print `compile` and `compileOnly` configurations classpath and assert that output contains proper classpath.

zolotov:
closing PR, implemented in master

PR #191, opened 2017-03-29
Fix test classpath

madisp:
Fixes test classpath by placing idea dependencies on compile (and
runtime) dependencies instead of compileOnly.

### DIFF
--- a/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
@@ -71,7 +71,7 @@ class IntelliJPlugin implements Plugin<Project> {
 
         if (JavaPlugin.hasProperty('COMPILE_ONLY_CONFIGURATION_NAME')) {
             project.configurations.getByName(JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME).extendsFrom idea, ideaPlugins
-            project.configurations.getByName(JavaPlugin.TEST_COMPILE_ONLY_CONFIGURATION_NAME).extendsFrom idea, ideaPlugins
+            project.configurations.getByName(JavaPlugin.TEST_COMPILE_CONFIGURATION_NAME).extendsFrom idea, ideaPlugins
         } else {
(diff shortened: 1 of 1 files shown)


src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
```diff
@@ -71,7 +71,7 @@ class IntelliJPlugin implements Plugin<Project> {
 
         if (JavaPlugin.hasProperty('COMPILE_ONLY_CONFIGURATION_NAME')) {
             project.configurations.getByName(JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME).extendsFrom idea, ideaPlugins
-            project.configurations.getByName(JavaPlugin.TEST_COMPILE_ONLY_CONFIGURATION_NAME).extendsFrom idea, ideaPlugins
+            project.configurations.getByName(JavaPlugin.TEST_COMPILE_CONFIGURATION_NAME).extendsFrom idea, ideaPlugins
```
zolotov: what's wrong with `TEST_COMPILE_ONLY_CONFIGURATION_NAME` here?

At least `ideaPlugins` should not be in test runtime

madisp: hmm, I had problems where it couldn't find JUnit on the classpath but I can't reproduce anymore. Maybe I had some other configuration issue at the same time? Feel free to reject the PR.

madisp: here are the classpath differences for a project that doesn't depend on any plugins: https://gist.github.com/madisp/ea0b0793363795a73f845b27ebe8c8d2

zolotov: Oh, I see. Then let's keep `idea` in classpath and exclude `ideaPlugins` from it.

Test Fixture loads all plugins from `idea.plugins.path` property and each plugin have its own ClassLoader to load classes from a corresponding directory, so having them in classpath look redundant at least.

zolotov:
closing PR, implemented in master

PR #214, opened 2017-07-17
Reset permissions for Rider remote dependency

citizenmatt:
The Rider distribution requires various files to have execute permissions set on Mac and Linux. The zip file doesn't maintain these permissions, and so they need to be reset after extracting the zip file. This PR will reset these permissions.


zolotov:
Thanks!

zolotov:
Will be available in 0.2.15 in an hour

PR #215, opened 2017-07-17
Add Rider Unity plugin to list of examples

citizenmatt:


zolotov:
Added example manually

PR #223, opened 2017-08-23
Use the new `classesDirs` API. Fixes #221

### LINKED ISSUE #221: Kotlin 1.1.4 break jar generation
I tried to use Kotlin 1.1.4 in my IntelliJ plugin, and the classes weren't copied into the plugin jar. 

When using 1.1.3, the classes are copied into it.
I can provide a sample if required but I think the issue is easy to reproduce.

[Kotlin Tracker Issue](https://youtrack.jetbrains.com/issue/KT-19737)

h0tk3y:


zolotov:
Added test, merged manually

PR #232, opened 2017-09-14
Document the possibility to use custom channels in dependency plugins

valich:
It looks like it's possible to depend on custom channel builds of plugins.
I've documented that in the respective section.


zolotov:
Thanks!

PR #243, opened 2017-10-24
Modify patching node method

ciscorucinski:
# Pull Request
## Summary
I generalized the patching mechanism so that a single method will handle most use-cases (in fact, it will handle 100% of current use-cases) while allowing for easy extension of the plugin to allow patching of other XML elements in each `plugin.xml` file.

## Implementation
This patching mechanism is accomplished by passing a Typed-`Map` that contains all the information of a single XML element (plus optional `break` logic) such as the **tag name**, **tag attributes**, and **tag data**.

Also. these changes are invisible to users of this plugin.

### Previous Patching Mechanism
The previous patching mechanism is only available for 5 tags...
1. `idea-version`
 2. `description` 
 3. `change-notes`
 4. `version`
 5. `id`

#### As Is
My current patching ability should allow you (JetBrains) to easily **add** the following 6 tags to the plugin capabilities...
 1. `name`
 2. `vendor` 
 3. `helpset`
 4. `resource-bundle`
 5. A single module dependency via `depends` 
    **OR**
 6. A single optional dependency via `depends optional="true"`).

I think 5 and 6 will only allow 1 total for both tags. Not 1 each; 1 total.

#### Slight Modification
With slight modifications, it should also be able to handle...
 1. `url` attribute for `idea-plugin` tag
 2. multiple `depends` tags

## Test Results
I only modified `PatchPluginXmlTask`, so I mostly only tested my modifications from `PatchPluginXmlSpec`
patchpluginxmlspec_results

## Limitations
My current implementation (minus slight changes) cannot allow for patching nested tags (that I am aware of). 

The following cannot be patched with my implementation without a second patching mechanism...
 1. `application-components`
 2. `project-components`
 3. `module-components`
 4. `actions`
 5. `extensionPoints` 
 6. `extensions`

### DIFF
--- a/src/main/groovy/org/jetbrains/intellij/tasks/PatchPluginXmlTask.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/tasks/PatchPluginXmlTask.groovy
@@ -138,12 +138,6 @@ class PatchPluginXmlTask extends ConventionTask {
         files.each { file ->
             def pluginXml = Utils.parseXml(file)
 
-//            patchSinceUntilBuild(getSinceBuild(), getUntilBuild(), pluginXml)
-//            patchNode("description", getPluginDescription(), pluginXml)
-//            patchNode("change-notes", getChangeNotes(), pluginXml)
-//            patchPluginVersion(getVersion(), pluginXml)
-//            patchNode("id", getPluginId(), pluginXml)
-
             patchNode(pluginXml, "idea-version", [
                     "@since-build" : getSinceBuild(),
                     "@until-build" : getUntilBuild()])
@@ -171,41 +165,6 @@ class PatchPluginXmlTask extends ConventionTask {
             }
         }
     }
-
-//    static void patchPluginVersion(String pluginVersion, Node pluginXml) {
-//        if (pluginVersion && pluginVersion != Project.DEFAULT_VERSION) {
-//            def version = pluginXml.version
-//            if (version) {
-//                version*.value = pluginVersion
-//            } else {
-//                pluginXml.children().add(0, new Node(null, 'version', pluginVersion))
-//            }
-//        }
-//    }
-//
-//    static void patchSinceUntilBuild(String sinceBuild, String untilBuild, Node pluginXml) {
-//        if (sinceBuild && untilBuild) {
-//            def ideaVersionTag = pluginXml.'idea-version'
-//            if (ideaVersionTag) {
-//                ideaVersionTag*.'@since-build' = sinceBuild
-//                ideaVersionTag*.'@until-build' = untilBuild
-//            } else {
-//                pluginXml.children().add(0, new Node(null, 'idea-version',
-//                        ['since-build': sinceBuild, 'until-build': untilBuild]))
-//            }
-//        }
-//    }
-//
-//    static void patchNode(String name, String value, Node pluginXml) {
-//        if (value != null) {
-//            def tag = pluginXml."$name"
-//            if(tag) {
-//                tag*.value = value
-//            } else {
-//                pluginXml.children().add(0, new Node(null, name, value))
-//            }
-//        }
-//    }
 
     static void patchNode(Node pluginXml, String tagName, Map<String, Object> values) {
         if (!values) return
--- a/src/main/groovy/org/jetbrains/intellij/tasks/PatchPluginXmlTask.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/tasks/PatchPluginXmlTask.groovy
@@ -18,7 +18,6 @@ class PatchPluginXmlTask extends ConventionTask {
     private Object pluginId
 
     private static final String TAG_DATA = 'value'
-    private static final String TAG_NAME = "_tag-name_"
     private static final String BREAK_LOGIC = "_break-logic_"
 
     @OutputDirectory
@@ -139,23 +138,15 @@ class PatchPluginXmlTask extends ConventionTask {
         files.each { file ->
             def pluginXml = Utils.parseXml(file)
 
-            patchNode(pluginXml, [
-                    (TAG_NAME) : "idea-version",
+            patchNode(pluginXml, "idea-version",  [
                     "@since-build" : getSinceBuild(),
                     "@until-build" : getUntilBuild()])
-            patchNode(pluginXml, [
-                    (TAG_NAME) : "description",
-                    (TAG_DATA) : getPluginDescription()])
-            patchNode(pluginXml, [
-                    (TAG_NAME) : "change-notes",
-                    (TAG_DATA) : getChangeNotes()])
-            patchNode(pluginXml, [
-                    (TAG_NAME) : "version",
+            patchNode(pluginXml, "description", [(TAG_DATA) : getPluginDescription()])
+            patchNode(pluginXml, "change-notes", [(TAG_DATA) : getChangeNotes()])
+            patchNode(pluginXml, "version", [
                     (TAG_DATA) : getVersion(),
                     (BREAK_LOGIC) : getVersion() == Project.DEFAULT_VERSION])
-            patchNode(pluginXml, [
-                    (TAG_NAME) : "id",
-                    (TAG_DATA) : getPluginId()])
+            patchNode(pluginXml, "id", [(TAG_DATA) : getPluginId()])
 
             def writer
             try {
@@ -172,20 +163,16 @@ class PatchPluginXmlTask extends ConventionTask {
         }
     }
 
-    static void patchNode(Node pluginXml, Map<String, Object> values) {
-        def tagName
+    static void patchNode(Node pluginXml, String tagName, Map<String, Object> values) {
         if (!values) return
         if (values.remove(BREAK_LOGIC)) return
-        if (!(tagName = values.remove(TAG_NAME))) return
         values.each { it ->
             if (!it.value) return
             def tag = pluginXml."$tagName"
             if (tag) {
                 tag*."$it.key" = it.value
             } else {
-                def value = it.key.startsWith("@") \
-                        ? ([(it.key[1..-1]) : it.value])
-                        : (it.value)
+                def value = it.key.startsWith("@") ? [(it.key[1..-1]) : it.value] : it.value
                 pluginXml.children().add(0, new Node(null, tagName, value))
             }
         }
--- a/src/main/groovy/org/jetbrains/intellij/tasks/PatchPluginXmlTask.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/tasks/PatchPluginXmlTask.groovy
@@ -17,6 +17,9 @@ class PatchPluginXmlTask extends ConventionTask {
     private Object changeNotes
     private Object pluginId
 
+    private static final String TAG_DATA = 'value'
+    private static final String BREAK_LOGIC = "_break-logic_"
+
     @OutputDirectory
     File getDestinationDir() {
         destinationDir != null ? project.file(destinationDir) : null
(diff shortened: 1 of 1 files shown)


zolotov:
Hi. Thank you for the pull request.

Could you please explain the motivation for these changes? The old approach looks more straightforward to me and it seems it's simpler in term of reading and further supporting.

As for advantages of the new approach, to be honest, I cannot imagine a scenario when its needed to update `vendor`, `helpset`, `resources-bundle` or module dependencies with gradle, can you?

ciscorucinski:
I went back to having the tag name part of the method signature and that allows me to call the patchNode method in one line for most of them. This makes the code seem less bloated.

Right now, the code you have has a bunch of duplicated code for patching XML elements. You have several dedicated methods that are only used for one xml element and nothing else. 

 What you have currently...
 1. Method for any tag with data only
 2. Method for a specific tag with 2 specific attributes
 3. Method for a specific tag with data only that has custom breaking logic

The current form of further support would result in more dedicated methods or by modifying the method signatures to allow for generalization. 

## Look at the difference

Right now my pull request allows for the easy addition of 6 additional xml elements. Lets compare adding those capabilites to my code and your code.

### Adding to my pull request
To add the 6 additional patching abilities that my change would allow, you only need to add the following 6 method calls. Nothing else is needed only; only these 6 method calls.

    patchNode(pluginXml, "name", [ (TAG_DATA) : getPluginName()])
    patchNode(pluginXml, "vendor", [ (TAG_DATA) : getVendorName(), "@url" : getVenderURL(), "@email" : getVenderEmail()])
    patchNode(pluginXml, "helpset", [ "@file" : getPluginHelpJarFile(), "@path" : getPathToHelpSetFileInJar()])
    patchNode(pluginXml, "resource-bundle", [ (TAG_DATA) : getResourceBundle()])
    patchNode(pluginXml, "depends", [ (TAG_DATA) : getModuleDependency()])
    patchNode(pluginXml, "depends", [ (TAG_DATA) : getOptionalDependency(), "@optional" : true, "@config-file" : getOptionalDependencyConfigFile()])

The map can place each key:value on a separate line for even easier readability (I personally believe). So the line count could be between 6 and 17 lines. **That's all for my changes; a minimum of 6 lines of code!**

### Adding to your code
To add the 6 additional patching abilities to your current code, you need to add the 6 method calls (like mine) plus several dedicated methods (about 10 additional similar lines each)

    patchNode("name", getPluginName(), pluginXml)
    patchVendor(getVendorName(), getVenderURL(), getVenderEmail(), pluginXml)
    patchHelpset(getPluginHelpJarFile(), getPathToHelpSetFileInJar(), pluginXml)
    patchResourceBundle("resource-bundle", getResourceBundle(), pluginXml)
    patchNode("depends", getModuleDependency(), pluginXml)
    patchOptionalDependency(getOptionalDependency(), getOptionalDependencyConfigFile(), pluginXml)

Already my changes appear easier to read to. I think many others would agree. However, you will need to add the following methods...

    static void patchVendor(String vendorName, String venderURL, String venderEmail, pluginXml) {
        // 9 additional similar lines 
    }

    static void patchHelpset(String helpJarFile, String helpSetPath, pluginXml) {
        // 8 additional similar lines
    }

    static void patchOptionalDependency(String dependency, String configFile(), pluginXml) {
        // 8 additional similar lines
    }

As you can clearly see, using your current method would result in about 40 lines of code. It is because of this and the duplicate code that I suggest this pull request.

ciscorucinski:
It would appear that none of the suggested 6 items change, and for many people that might be the case; however, as you will see, it isn't unchanged for everyone and there appear to be use-cases that you have not thought of yet.

#### Vendor Support
While in university, my email I wanted people to get in contact with was my university email and my url was my Facebook account. After university, both of my email and url was with Facebook. Now, I use my personal email and my github page. Later on, I might make dedicated email addresses for each plugin and my own personal blog (github blog?).

I could create some Gradle properties for these on my own system, and just redeploy my plugins with the new values. Or I can go one by one and update my plugins.

Also, I could create a template for all new plugins that have my information already and Gradle will automatically update that information in plugin.xml

#### ID Support
This is something that JetBrains advises developers not to change, yet this is available and has consequences some developers might not know about.

#### Name Support
Entities all over the world, of all kinds, change their names occasionally. As for plugins, some have new goals and aspirations for their plugin and find a new name more fitting. Some have [legal problems](https://intellij-support.jetbrains.com/hc/en-us/community/posts/206764549-How-do-I-change-the-plugin-name-on-the-JetBrains-Plugin-Site-) and are required to change their name.

This is something I am considering for my plugin. It is a color scheme plugin that I intend to update and go beyond its original use-case of just a color scheme. If I want the name "ChroMATERIAL" to be associated with the color scheme aspect only, then I will be forced to change the name of the plugin and indicate the plugin uses ChroMATERIAL.

I am sure there are a ton more examples that were never thought of

#### Helpset and Resource-Bundle Support
I'll give you a point here, these are not really needed, but...

If added, it could allow for documentation of these features to be combined in this central location. I don't have to search all over the internet to figure out how to use it.  For example, if I don't know which module dependency to use, then I have to search and find the page only about that. Then if I need help with Helpset, I have to search and find a different page that is disconnected from the first. This is the same for Resource-bundle and many others. Having all of these supported allows for documentation to be grouped together. 

I have seen both users and JetBrain employees stating about documentation not being the best or it not even existing. This was also a reason why I updated the README of this plugin as I felt it could have been better.

#### Dependency Support
This is where I think this plugin can really shine. I can talk about that later if needed. 

A lot of developers will add new features and abilities to their plugins, and that means that their dependencies (possibly even module dependencies) will change. This will be the case for my own "ChroMATERIAL" color scheme plugin as I turn it into a real plugin that has color scheme support.

This is also the case when users of a plugin find or ask for changes in a plugin. I did just that with the "AdvancedExpressionFolding" java-based plugin from a JetBrain's employee. It only worked on IntelliJ, but I manually installed the plugin in Android Studio and it appeared to work without problems. So I [suggested the plugin have Android Studio Support](https://github.com/cheptsov/AdvancedExpressionFolding/issues/30) which meant the module dependency had to change.

ciscorucinski:
Also, we use plugin.xml because that is how it has always been. I get that and that won't change.  

I come from the idea that...as long as the Gradle DSL is sufficient, a Gradle-only patched plugin.xml is far superior to anything else. I don't think about it as patching plugin.xml as much as **Gradle being a facade to plugin.xml**. I also understand that would take a good while to implement.

zolotov:
Regarding generalization, I prefer adding 33 more lines of simple and flexible imperative code instead of figuring out every time what `BREAK_LOGIC` is for and trying to parse or write something like `patchNode(pluginXml, "helpset", [ "@file" : getPluginHelpJarFile(), "@path" : getPathToHelpSetFileInJar()])`.

I doubt any of these fields would ever be added. No, Gradle is not a facade for plugin.xml. `plugin.xml` is and will be the entry point for the plugin and the place that should be edited. Gradle is a build tool, nothing more, it can help you to modify dynamic fields line `idea-version` or `description` because Gradle knows more about it or it was a **real** usecase in plugins development.

As for 6 items to change.

- `vendor` usually includes link to plugin's url to documentation or to github, so you cannot generalize it. If it's a plugin made by company, I don't think company's url will be changed. Also, I'd check how many plugins people usually have at plugins.jetbrains.com, and I'm not sure it's really the case when one wants to update vendor for 10 plugin in a row.
> Also, I could create a template for all new plugins that have my information already and Gradle will automatically update that information in plugin.xml

You can make a template for `plugin.xml` as well.

- `ID` is editable from Gradle right now. And I'd deleted this ability if I was sure that it won't break anything.

- `name` is also very plugin-specific and static information, no need to change it from Gradle.

- Sorry, but I think you misinterpret `helpset` and `resource-bundle` tags. As far as I understand, they cannot be combined in a central location. They are paths to files inside your plugin. So they also very plugin-specific and it doesn't make any sense to generate them during building.

- The same for `dependency`.  Of course, they changed, but there is no need to change them dynamically during the build.

zolotov:
@halirutan, could you please, as a voter, tell your use cases regarging these changes? thanks!

halirutan:
@zolotov It's not so much that I *need* this, but it would be nice to have all the information about the plugin in one place. If this would be possible, I would throw all things except the extensions, actions, etc out of my `plugin.xml`. I would include `id`, `name`, `category`, and `vendor` in my `build.gradle` not because I want to change them every day, but because it's a clearer layout for me.

In addition to that, I looked over the changes @ciscorucinski made and I like that he generalized the patching of XML elements. Even if the direct use-cases for this hit only some developers, I think it is an improvement for the code. But this is of course only my view from looking on the diffs and I have never really looked over the whole codebase.

ciscorucinski:
@halirutan is right in that it will only affect some developers, but it is nice to consolidate information and help developers have a more enjoyable experience (including those outside of JetBrains and outside of companies).

@zolotov To be clear, I only generalized **the code** for patching xml elements. Can't generalize items within plugin.xml. Not what I was suggesting at all. Also, my code is your imperative code. There is almost no difference. I took your code, renamed variables, found similarities and differences to refactor code, and looped through map values.

### Explaining Code
#### Line 173
My key discovery was `tag*."$it.key" = it.value` and using a **map**. When `tag*."$it.key"` is evaluated, it can be similar to one of the following...
 * `tag*.value` will set the tag data (`<tag> tag data </tag>`). 
 * `tag*.@example` will give the tag an attribute called `example` (`<tag example=""> ...`)
 
#### That means
 * At most 1 key can have the name `value`. 
 * All attributes must have a key name that starts with `@`. 
 * The key `BREAK_LOGIC` will have a boolean value. If it evaluates to `false` or `null` then that tag is patched otherwise it is skipped completely. 

If you are confused at any of my code, then I can explain it further!

zolotov:
Thank you :) I am able to figure out what the code does.

The point is I won't remember what `BREAK_LOGIC` means tomorrow and I don't want to spend time to understand the code again every time I change this class. I should have a good reason for that, reducing line number in unlikely future is not the good reason to me.

What do you think about making the solution less DSL'ish and getting rid of magic constants?

I believe it can be done by 
- pulling BREAK_LOGIC part up to the method's client
- getting rid of '@' in map's keys and moving this logic into the method
- migrate TAG_CONTENT part to a new method's parameter

ciscorucinski:
## Update
The following has been updated locally but not pushed. I took your recommendations and applied them as best as possible. 

If these changes seem good, then I will push them. Otherwise, please comment on other changes you would like to see.

### Changes
#### Directly Solves Points
 * Created an auxiliary method to separate out tags that need custom logic to decide if they should be skipped or not. _I think this is suitable but unsure if exactly what you wanted._
 * Moved tag content out of the map and into the parameter list.
 * Moved logic of attributes needing `@` into the patching method

#### Possible Undesired Changes
 * If a tag has attributes but no content, then the patching method call will include a `null` value, such as with the tag `idea-version`
#### Other Changes
 * Renamed patching method from `patchNode()` to `patchTag`. _I believe a Node in XML can refer to a tag, attribute, or content._
 * Renamed some parameters. _Looks better with parameter hinting._

### Code
#### Calling Patching Method(s)

Here is what the calling code will look like with my new code.

```
    patchTag(pluginXml, "idea-version", null, [
            "since-build": getSinceBuild(),
            "until-build": getUntilBuild()])
    patchTag(pluginXml, "description", getPluginDescription())
    patchTag(pluginXml, "change-notes", getChangeNotes())
    patchTagIf(getVersion() != Project.DEFAULT_VERSION,
            pluginXml, "version", getVersion())
    patchTag(pluginXml, "id", getPluginId())
```

 #### Modified Patching Method

##### Pseudocode
 * Skip tag If name or plugin are not provided
 * Loop through attributes, prepend `@` to key, and add to a new nodes map
 * Loop through nodes (tag content and attributes) like before. 

```
    static void patchTag(Node pluginXml, String name, String content = null, Map<String, Object> attributes = [:]) {
        if (!name || !pluginXml) return
        def tagNodes = content != null ? [value: content] : [:]
        for (attribute in attributes) {
            tagNodes += [("@$attribute.key" as String) : attribute.value]
        }
        tagNodes.each {
            if (!it.value) return
            def tag = pluginXml."$name"
            if (tag) {
                tag*."$it.key" = it.value
            } else {
                def value = it.key.startsWith("@") ? [(it.key[1..-1]) : it.value] : it.value
                pluginXml.children().add(0, new Node(null, name, value))
            }
        }
    }
```

#### Auxiliary Patching Method
This method is to remove BREAK_LOGIC magic constant. 

It only cares about the boolean value. All other values get passed along without validation to the patching method.
```
    static void patchTagIf(boolean shouldPatch, Node pluginXml, String name, String content = null, Map<String, Object> attributes = [:]) {
        if (shouldPatch) patchTag(pluginXml, name, content, attributes)
    }
```

zolotov:
I don't think that auxiliary patching method is needed, but it's ok to have it.

The rest looks good to me. Please push the changes.

zolotov:
Merged, thanks

PR #246, opened 2017-10-29
Add Bamboo Soy plugin to examples list

yorlov:


zolotov:
README.md is generated, use `docs/` instead

PR #270, opened 2018-03-20
Recognize new names for kotlin-stdlib as IDEA dependency

yole:
Ref: https://github.com/JetBrains/intellij-community/commit/202e78d33

### DIFF
--- a/src/main/groovy/org/jetbrains/intellij/dependency/IdeaDependencyManager.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/dependency/IdeaDependencyManager.groovy
@@ -295,12 +295,16 @@ class IdeaDependencyManager {
         def configurations = project.configurations
         def closure = {
             if ("org.jetbrains.kotlin" == it.group) {
-                return "kotlin-runtime" == it.name || it.name.startsWith('kotlin-stdlib') || "kotlin-reflect" == it.name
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/jetbrains/intellij/dependency/IdeaDependencyManager.groovy
```diff
@@ -295,13 +295,17 @@ class IdeaDependencyManager {
...
+                return isKotlinRuntime(it.name)
             }
             return false
         }
         return configurations.getByName(JavaPlugin.RUNTIME_CONFIGURATION_NAME).getAllDependencies().find(closure) ||
                 configurations.getByName(JavaPlugin.COMPILE_CONFIGURATION_NAME).getAllDependencies().find(closure)
     }
+
+    static boolean isKotlinRuntime(name) {
+        return "kotlin-runtime" == name || name.startsWith('kotlin-stdlib') || "kotlin-reflect" == name
```
zolotov: the first usage (in `IdeaDependency`) matches file names, `kotlin-rutime == name` will be always false for `kotlin-runtime.jar` file. I think we can do something like `file.name - '.jar'` in the first usage

zolotov:
I changed it. Will publish 0.2.20 and 0.3-SNAPSHOT in an hour

PR #286, opened 2018-05-23
Fix example project's link

oldergod:
cc @AlecStrong


JakeWharton:
Why don't we just link to the repo?

AlecStrong:
i believe it was so that the linked directory had the gradle file which applied this plugin directly

JakeWharton:
Yeah that makes sense. But it should be fairly easy to discover based on the folder name. Plus few others link to anything but the enclosing repo.

zolotov:
README.md is generated. Could you please edit `src/docs/examples.md`, regenerate readme with `documentr` task and push both files? Thanks

oldergod:
Done.

PR #291, opened 2018-06-08
fix compatibility for SDK 182

Eskibear:
In SDK 182, `jgoodies-forms.jar` has been renamed to `forms-1.1-preview.jar`.

image

Relate issue: https://github.com/JetBrains/gradle-intellij-plugin/issues/290


zolotov:
fixed with wildcard as a version

PR #306, opened 2018-07-26
Update Kotlin

jonnyzzz:
Let's use Kotlin 1.2.51, which is included into IJ 2018.2


zolotov:
Hi! Any reason for this? Btw, in the example `2016.3.1` is used.

jonnyzzz:
Good point. Anyways, all users will likely to upgrade it at first. Will it
fail?

zolotov:
Upgraded manually

PR #308, opened 2018-07-27
Relax ivy artifact pattern

EddieRingle:
Using this plugin with Gradle 4.9 to build a project
that included the AsciiDoctor IDEA plugin was failing
due to Gradle being unable to find AsciiDoctor's
classes directory. Relaxing the pattern to make the
extension part (particularly the period preceding it)
optional fixes the issue.

Signed-off-by: Eddie Ringle <eddie@ringle.io>

### DIFF
--- a/src/main/groovy/org/jetbrains/intellij/dependency/PluginDependencyManager.groovy
+++ b/src/main/groovy/org/jetbrains/intellij/dependency/PluginDependencyManager.groovy
@@ -79,7 +79,7 @@ class PluginDependencyManager {
         project.repositories.ivy { repo ->
             repo.ivyPattern("$cacheDirectoryPath/[module]-[revision].[ext]") // ivy xml
             repo.artifactPattern("$ideaDependency.classes/plugins/[module]/[artifact].[ext]") // builtin plugins
-            repo.artifactPattern("$cacheDirectoryPath/[module]-[revision]/[artifact].[ext]") // external plugins
+            repo.artifactPattern("$cacheDirectoryPath/[module]-[revision]/[artifact](.[ext])") // external plugins
             pluginSources.each {
                 repo.artifactPattern("$it/[artifact].[ext]") // local plugins
             }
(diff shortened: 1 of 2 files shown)


EddieRingle:
Note that I had the new test set `instrumentCode` to false. I was running into an issue where the instrumentation task was failing due to being unable to find some of the shaded ASM classes, despite those jars from the IDE being included in the classpath.

zolotov:
I would avoid relaxing patterns. Since Gradle 4.<something> the first matched repository will be used for resolving a dependency, and it won't fall back to the next repositories in case of missing dependency. So relaxing pattern leads to more probable failed builds.

What exactly the problem with AsciiDoctor IDEA classes directory? Is it in irregular place or something?

EddieRingle:
AFAIK this only relaxes the artifact matching pattern after finding a repository, but I could be wrong.

The dependencies in my gradle cache directory look as follows:

screenshot from 2018-07-30 11-25-21

It's been a few days since I looked at this so I apologize as I've forgotten a bit of the specifics, but I believe it was coming down to how Gradle was building the path to the dependency based on the pattern. If I get some time today I'll try to look back through things to find more details, if you'd like.

Without making the extension part optional, the test ends in a failure (the same one I ran into trying to build my plugin with AsciiDoc as a dependency):

```
Could not resolve all files for configuration ':compileClasspath'.
> Could not find asciidoctor/classes/ (org.jetbrains.plugins:org.asciidoctor.intellij.asciidoc:0.20.6).
> Could not find asciidoctor/META-INF/ (org.jetbrains.plugins:org.asciidoctor.intellij.asciidoc:0.20.6).
```

zolotov:
> AFAIK this only relaxes the artifact matching pattern after finding a repository, but I could be wrong.

It does relax the artifact matching pattern, and as soon as it matched, gradle won't look for the dependency in other repositories even if the dependency is missing in the current one. See https://github.com/gradle/gradle/issues/3330 for details.

PR #318, opened 2018-09-19
Fix markerFile logic

nd:
We were looking for build.txt inside parent of unzipped dir, didn't
find it there, and didn't write anything into markerFile. As a result
same file was unzipped many times.


zolotov:
Fixed in a bit different way. Thanks for the catch

Issue #303, opened 2018-07-25
Can't run plugin on Windows due to long java exec command


satamas:
After I updated IntelliJ version to 2018.2 it wasn't possible to build plugin due to such errors

> CreateProcess error=206, The filename or extension is too long

This happens because all Idea Communtiy jars are added to classpath one by one and this is too much for windows. As a workaround I've used [this plugin](https://github.com/viswaramamoorthy/gradle-util-plugins) that  creates a manifest jar for jar files in the classpath. Can gradle plugin do this by default?

zolotov:
As a workaround use `intellij.ideaDependencyCachePath = project.buildDir`, so path to jar files won't be too long

PR #330, opened 2018-10-17
Override Rider SDK cache directory on Windows only

auduchinok:
Rider SDK cache directory is being overridden to prevent long path exceptions on Windows. However, there's no point to override it on macOS and Linux.
Reusing Gradle caches, the default behaviour, helps when having multiple working copies or plugins using the SDK.


citizenmatt:
Rider has an additional problem on top of #303, that the C# nuget packages that are in the SDK can't be referenced on Windows because the paths are too long. So fixing #303 still wouldn't help.

PR #362, opened 2019-01-29
Fix broken link

funivan:


zolotov:
readme is generated, you should change docs/* files instead and regenerated readme.md with `documentr` task

PR #371, opened 2019-03-07
Update README.md

yrusskih:


zolotov:
readme is generated, it edited the sources and regenerated it in master

PR #383, opened 2019-04-04
Remove gcloud plugin repo from list of examples

matthewmichihara:
The gcloud plugin repo is now just for docs and issues.


zolotov:
thanks

PR #384, opened 2019-04-10
enable caches, improve tests and docs

artspb:
The caches will work from the first attempt when [this issue](https://youtrack.jetbrains.com/issue/IDEA-210505) will be fixed.

### DIFF
--- a/src/test/groovy/org/jetbrains/intellij/SearchableOptionsSpecBase.groovy
+++ b/src/test/groovy/org/jetbrains/intellij/SearchableOptionsSpecBase.groovy
@@ -1,0 +1,73 @@
+package org.jetbrains.intellij
+
+class SearchableOptionsSpecBase extends IntelliJPluginSpecBase {
+
+    @Override
+    String getIntellijVersion() {
+        return '2019.1'
+    }
+
+    protected static String getPluginXmlWithSearchableConfigurable() {
+        """<idea-plugin version="2">
+               <name>PluginName</name>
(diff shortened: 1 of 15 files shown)


src/test/groovy/org/jetbrains/intellij/SearchableOptionsSpecBase.groovy
```diff
@@ -0,0 +1,73 @@
+package org.jetbrains.intellij
+
+class SearchableOptionsSpecBase extends IntelliJPluginSpecBase {
+
+    @Override
+    String getIntellijVersion() {
+        return '2019.1'
```
zolotov: it's better not to introduce the new version of IDE in tests in order not to increase testing time. If you need 2019.1, let's replace all usages of 2017.2.5 to it

artspb: Tests for searchable options use both 2017.2.5 and 2019.1. The former is needed to check that steps are disabled for old versions and the later is used to ensure that they actually work. Should I drop negative tests then?

zolotov: there is another old version, 14.1.4. you can use it or migrate them to 2017.2.5 and 2017.2.5 to 2019.1.

artspb: Done. Took a while, though...

PR #390, opened 2019-04-29
Possibility to turn off building searchable options

yole:
If I know my plugin does not contain any configurables, I can speed
up its build by turning off the searchable options builder.


zolotov:
I think it can be disabled as task: buildSearchableOptions.enabled=false?

zolotov:
updated FAQ: https://github.com/JetBrains/gradle-intellij-plugin/blob/master/FAQ.md#how-to-disable-building-searchable-options

PR #398, opened 2019-06-04
Added a tip about debugging gradle-intellij-plugin

vlasovskikh:


zolotov:
Sorry, Readme is generated. The text should be part of "src/docs/..." files. Then please run `./gradlew :documentr` task

zolotov:
thanks!

Issue #421, opened 2019-08-10
The build result accumulates searchableOptions.jar from multiple versions


ForNeVeR:
1. Create a new IDEA project using gradle-intellij-plugin:
   ```gradle
   plugins {
       id 'org.jetbrains.intellij' version '0.4.10'
   }
   
   group 'mytest'
   version '1.0-SNAPSHOT'
   
   // See https://github.com/JetBrains/gradle-intellij-plugin/
   intellij {
       version '2019.2'
   }
   patchPluginXml {
       changeNotes """
         Add change notes here.<br>
         <em>most HTML tags may be used</em>"""
   }
   ```
2. Build the plugin:
   ```console
   $ ./gradlew buildPlugin
   ```
3. Update the plugin version in `build.gradle`:
   ```patch
   - version '1.0-SNAPSHOT'
   + version '1.0'
   ```
4. Build the plugin again:
   ```console
   $ ./gradlew buildPlugin
   ```
5. Inspect the resulting `build/distributions/*-1.0.zip`

**Expected result**: the build result should include the latest searchable options artifact, namely `searchableOptions-1.0.jar`.
**Observed result**: both `searchableOptions-1.0-SNAPSHOT.jar` and `searchableOptions-1.0.jar` are included. It leads to unnecessary jar pollution. Some of the plugins (e.g. [codota](https://plugins.jetbrains.com/plugin/7638-codota-/versions)) have this problem already.

PR #427, opened 2019-09-02
#421 include only relevant searchableOptions.jar

artspb:


src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
```diff
@@ -507,12 +507,12 @@ class IntelliJPlugin implements Plugin<Project> {
...
             description = "Bundles the project as a distribution."
             group = GROUP_NAME
-            from({
-                [
-                        "${prepareSandboxTask.getDestinationDir()}/${prepareSandboxTask.getPluginName()}",
-                        jarSearchableOptionsTask.getDestinationDir()
-                ]
-            })
+            from { "${prepareSandboxTask.getDestinationDir()}/${prepareSandboxTask.getPluginName()}" }
+            from({ jarSearchableOptionsTask.getDestinationDir() }) {
```
zolotov: can't you use just `from { "${jarSearchableOptionsTask.getDestinationDir()}/${jarSearchableOptionsTask.getArchiveName()}"}`?

artspb:
Nope. The structure is all follows: `pluginName/lib/*.jar`. If I do this
the file is copied to `pluginName` but I need it in `pluginName/lib`. I'll
try it again tomorrow but I think it didn't work or maybe I just forgot
braces :)

src/main/groovy/org/jetbrains/intellij/IntelliJPlugin.groovy
artspb: Nope. The structure is all follows: `pluginName/lib/*.jar`. If I do this
the file is copied to `pluginName` but I need it in `pluginName/lib`. I'll
try it again tomorrow but I think it didn't work or maybe I just forgot
braces :)

artspb: Confirm, your suggestion leads to `pluginName/searchableOptions.jar` instead of `pluginName/lib/searchableOptions.jar`.

zolotov: can you adjust `to` then?
it's just quite weird that we need to build complex include rule when actually we know the exact path of a file to copy

artspb: Can we be sure that `${prepareSandboxTask.getDestinationDir()}/${prepareSandboxTask.getPluginName()}` always contains only `lib` directory? In this case, we can adjust `into` as it's used by both paths.

zolotov: yes, `lib` directory might be missing if you have no sources in the project. But in this case, searchable options will be also missing.

zolotov:
thanks!

PR #507, opened 2020-06-23
Provide detailed exception message in PublishTask

hsz:


zolotov:
Hm, what's the point of putting exception stack trace into the message?

PR #518, opened 2020-07-20
Mention T4 plugin in README

kirillgla:
I've been developing this plugin for like a year, and it's been bundled in Rider for quite a while, so I believe it deserves being mentioned


zolotov:
Examples part is not a hall of fame, it's supposed to bring some information  for users about how to use the plugin. Does T4 bring something new in terms of build configuration comparing to F# plugin? Please mention those points so users could easier filter the plugin they're interested in.

Also, please not that README.me is generated file (`:documentr` task). The source is `src/docs/examples.md`.

kirillgla:
Fair enough. The build configuration in T4 plugin is not too different from the one in F# plugin

PR #525, opened 2020-08-31
Add option to easily disable auto-reload of plugins

piotrtomiak:


src/main/groovy/org/jetbrains/intellij/tasks/RunIdeBase.groovy
```diff
@@ -236,7 +252,10 @@ abstract class RunIdeBase extends JavaExec {
         }
         systemPropertyIfNotDefined("idea.classpath.index.enabled", false, userDefinedSystemProperties)
         systemPropertyIfNotDefined("idea.is.internal", true, userDefinedSystemProperties)
-        systemPropertyIfNotDefined("idea.auto.reload.plugins", true, userDefinedSystemProperties)
+
+        if (!userDefinedSystemProperties.containsKey("idea.auto.reload.plugins") && isAutoReloadPlugins()) {
```
zolotov, line 256: shouldn't we set `idea.auto.reload.plugins=false` if system property wasn't set and autoReloadPlugins=false?

README.md
```diff
@@ -1,10 +1,10 @@
 <a name="documentr_top"></a>[![official JetBrains project](https://jb.gg/badges/official.svg)](https://confluence.jetbrains.com/display/ALL/JetBrains+on+GitHub) [![Build Status](https://api.cirrus-ci.com/github/JetBrains/gradle-intellij-plugin.svg)](https://cirrus-ci.com/github/JetBrains/gradle-intellij-plugin)
  [![Twitter Follow](https://img.shields.io/twitter/follow/JBPlatform?style=flat)](https://twitter.com/JBPlatform/)
- [![Gradle Plugin Release](https://img.shields.io/badge/gradle%20plugin-0.4.21-blue.svg)](https://plugins.gradle.org/plugin/org.jetbrains.intellij) [![GitHub Release](https://img.shields.io/github/release/jetbrains/gradle-intellij-plugin.svg)](https://github.com/jetbrains/gradle-intellij-plugin/releases) 
+ [![Gradle Plugin Release](https://img.shields.io/badge/gradle%20plugin-0.4.22-blue.svg)](https://plugins.gradle.org/plugin/org.jetbrains.intellij) [![GitHub Release](https://img.shields.io/github/release/jetbrains/gradle-intellij-plugin.svg)](https://github.com/jetbrains/gradle-intellij-plugin/releases)
```
zolotov, line 3: readme is generated by `:documetr` task, versions will be fixed automatically, sources are in `docs/` directory

zolotov (changes requested):

src/main/groovy/org/jetbrains/intellij/tasks/RunIdeBase.groovy
piotrtomiak: If we pass the property, even as false, in 2020.1 and 2020.1.1 it will get enabled anyway (in IJ code there is check for property presence, not it's value), so not passing anything should fix the problem. This changes the behavior for newer versions as well, but I don't seen why would we want to pass the property as `false`?

README.md
piotrtomiak: ok, is this the only place, which is automatically updated, or are there are places which I've unnecessarily updated?

zolotov: all places will be updated

src/main/groovy/org/jetbrains/intellij/tasks/RunIdeBase.groovy
zolotov: >  but I don't seen why would we want to pass the property as false?

It looks more consistent to me.

Ok, let's keep it in this way

README.md
ForNeVeR: You have to repeat this for almost every new project contributor. Could we add a comment mentioning `:documetr` into each generated document, or something like that? AFAIK, Markdown should properly support HTML comments.

piotrtomiak:
I've pushed the change for declaring the plugin in the build file as well. It allows for much better recognition of the plugin when included locally. Now, one only needs to add to `settings.gradle.kts`:
```kotlin
includeBuild(".../gradle-intellij-plugin")
```
and include plugin using id (without a version) in `build.gradle.kts`:
```kotlin
plugins {
    id("org.jetbrains.intellij")
}
```

README.md
piotrtomiak: All of the version changes are removed.

zolotov: @piotrtomiak `docs` directory is still untouched, the entire README will be overwritten and changes you've made will be lost. Please make changes in `docs` dir and run `:documentr` task.

@ForNeVeR only README.md is generated. Let's try

piotrtomiak: @zolotov sorry, I missed the point. Should be fine now.

zolotov:
thank you very much, merged, will publish soon

PR #538, opened 2020-09-23
Update maven plugin to maven-publish

akainth015:
A bit of the code has also been updated to suppress spell check in dependency coordinates and change the usage of deprecated variables.


zolotov:
please, do not suppress spelling inspection. If a word it's complaining about is correct, add it to the dictionary

zolotov:
thanks for the change :)

PR #558, opened 2020-11-01
Add prepareTestingSandbox as an input to tests

### LINKED ISSUE #541: tasks.test is up-to-date even if the output of tasks.prepareTestingSandbox has changed
Inspired by the [FAQ](https://github.com/JetBrains/gradle-intellij-plugin/blob/master/FAQ.md#how-do-i-add-my-a-custom-file-inside-plugin-distribution) I have some code in build.gradle.kts which installs custom artifacts in my plugin home directory:
```kotlin
tasks.prepareTestingSandbox {
    from(customPluginArtifacts) {
        into(intellij.pluginName)
    }
}
```

The problem is that the outputs of the `prepareTestingSandbox` task are not modeled as inputs to the `test` task. I.e., in my example, if the `customPluginArtifacts` change, then the `test` task remains up-to-date and is skipped.

It was easy for me to fix it by adding this line:
```kotlin
tasks.test {
    inputs.files(tasks.prepareTestingSandbox)
}
```

However, I wonder if the gradle-intellij-plugin should always add this dependency automatically. What do you think?

gharrma:
This ensures that tests are out-of-date when the contents
of the test sandbox change.

Fixes #541


zolotov:
thanks!

PR #564, opened 2020-11-10
RunPluginVerifierTask: support for offline mode

hsz:
RunPluginVerifierTask: support for offline flag implemented with https://github.com/JetBrains/intellij-plugin-verifier/pull/58


zolotov:
Can we reuse gradle's offline flag instead of extending the API?

https://docs.gradle.org/current/javadoc/org/gradle/StartParameter.html#isOffline--

hsz:
That's a good idea!
I've implemented it already + tests.

PR #588, opened 2021-01-26
Migration of the Groovy tests to the Kotlin language

hsz:


build.gradle.kts
```diff
@@ -8,9 +11,14 @@ plugins {
...
-    tasks.withType<GroovyCompile> {
-        sourceCompatibility = "1.7"
-        targetCompatibility = "1.7"
+    tasks {
+        withType<GroovyCompile> {
+            sourceCompatibility = "1.7"
+            targetCompatibility = "1.7"
+        }
+        withType<KotlinCompile> {
+            kotlinOptions.jvmTarget = "1.8"
```
zolotov, line 20: 1.7 was chosen as it minimal working version for Gradle, change it only if it was changed there too

src/test/kotlin/org/jetbrains/intellij/DownloadIntelliJSpec.kt
```diff
@@ -0,0 +1,112 @@
+package org.jetbrains.intellij
+
+import org.gradle.api.plugins.BasePlugin
+import java.io.File
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+
+class DownloadIntelliJSpec : IntelliJPluginSpecBase() {
```
zolotov, line 9: what happened with Stepwise annotation?

zolotov (changes requested):
I'd like not to loose history of files. It contains many explanations of different hacks and reasons. Could you try to preserve it?

src/test/kotlin/org/jetbrains/intellij/IntelliJPluginSpecBase.kt
```diff
@@ -0,0 +1,181 @@
...
+            .withArguments(*tasks, "--stacktrace")
+
+    fun tasks(groupName: String): List<String> = build(ProjectInternal.TASKS_TASK).output.lines().run {
+        val start = indexOfFirst { it.equals("$groupName tasks", ignoreCase = true) } + 2
+        drop(start).takeWhile(String::isNotEmpty).map { it.substringBefore(' ') }
+    }
+
+
+    protected fun directory(path: String) = File(dir, path).apply { mkdirs() }
+//
```
zolotov, line 108: ?

build.gradle.kts
hsz: I get it, but Scala – in the end – will be completely removed anyway.

src/test/kotlin/org/jetbrains/intellij/IntelliJPluginSpecBase.kt
hsz: Missing leftover.

PR #593, opened 2021-02-02
Fix GoLand support

pleomaxx3002:


YannCebron:
we currently recommend this approach https://plugins.jetbrains.com/docs/intellij/goland.html#configuring-plugin-projects-targeting-goland  - what needs to be updated? is it correct that this patch requires targeting Goland 2019.3.1 and newer only?

pleomaxx3002:
This patch provides an option to download Goland which simplifies the setup.  It differs from recommended approach and just mimics the similar code for other IDE-s. 
I have tried it only for the latest version 2020.3.1

zolotov:
@artspb should the recommended way changed since we have goland distribution on maven?

artspb:
AFAIR we've started uploading GoLand to Maven since 2020.2. I think it makes sense to change the recommended approach and update the documentation accordingly. @pleomaxx3002 Could you please check that it works well with 2020.2?

pleomaxx3002:
This brunch is able to download and works properly with configuration
```properties
intellij {
    pluginName = "GoLandPlugin"
    version = "2020.2"
    type = "GO"
}
```

zolotov:
@artspb ok, thanks

@pleomaxx3002 could you, please, also change `docs/` sources and regenerated `README.md` with `:documentr` task? Otherwise, Readme will be overwritten on the next generating.

PR #599, opened 2021-02-17
JbrResolver: Use `jbr_jcef` prefix for JBR11 in version 1319.6+ (ref. https://confluence.jetbrains.com/pages/viewpage.action?pageId=206537703)

hsz:


ahus1:
@hsz - until this change has been released in the gradle plugin it is hard for me to test my plugin that builds on top of JCEF that with the latest EAP. Is there a workaround to run the latest EAP with the right JDK? Thanks!

ahus1:
(comment above is obsolete as 0.7.0 has been release by @zolotov - thanks a lot!)

zolotov:
0.7.1 would be better, the bug was revealed right after 0.7.0 publication

PR #634, opened 2021-04-28
Add sign plugin task

hsz:


src/main/kotlin/org/jetbrains/intellij/tasks/SignPluginTask.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.gradle.api.internal.ConventionTask
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.*
+import org.jetbrains.zip.signer.signer.CertificateUtils
+import org.jetbrains.zip.signer.signer.PrivateKeyUtils
+import org.jetbrains.zip.signer.signer.PublicKeyUtils
+import org.jetbrains.zip.signer.signing.DefaultSignatureProvider
+import org.jetbrains.zip.signer.signing.ZipSigner
+
+open class SignPluginTask : ConventionTask() {
```
YannCebron, line 14: can we have kdoc on all properties? so it's readily available during build file editing in IDE

src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
```diff
@@ -701,6 +720,11 @@ open class IntelliJPlugin : Plugin<Project> {
 
             task.dependsOn(buildPluginTask)
             task.dependsOn(verifyPluginTask)
+            task.dependsOn(signPluginTask)
+        }
+
+        project.afterEvaluate {
```
zolotov, line 726: Why is `afterEvaluate` required here? it's a bit suspicious.

hsz:
-> https://github.com/JetBrains/gradle-intellij-plugin/pull/635

src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
satamas: If the sign plugin task is enabled we should deploy the signed artifact. In another case, we should deploy an unsigned artifact. I don't know a better way to do this

zolotov: my question is why it's under `afterEvaluate`. Cannot it be unwrapped?

satamas: Because the value of the `enabled` property is known only after evaluation as far as I understand

zolotov: I see. I don't think this is the good way to handle that. Tasks state being manipulated after evaluation means that  user's data or some other plugins data might be lost. 

I think `enabled` flag is just wrong. `Task` already have one, duplicating in in an extension doesn't make much sense but brings some tricky questions about evaluation order and data loss.

PR #653, opened 2021-05-05
Configuration cache

hsz:


src/main/kotlin/org/jetbrains/intellij/utils.kt
```diff
@@ -131,27 +133,19 @@ fun Node.get(name: String) = childNodes.asSequence().find { it.nodeName == name
...
+fun File.isJar() = toPath().isJar()
 
-fun isZipFile(file: File) = file.toPath().isZip()
+fun File.isZip() = toPath().isZip()
 
-fun collectJars(directory: File, filter: Predicate<File>): Collection<File> = when {
-    !directory.isDirectory -> emptyList()
-    else -> FileUtils.listFiles(directory, object : AbstractFileFilter() {
-        override fun accept(file: File) = isJarFile(file) && filter.test(file)
+fun File.collectJars(filter: Predicate<File> = Predicate { true }): Collection<File> = when {
```
zolotov, line 140: I don’t think it’s good to have`collectJars` in completion for `File` in any part of the program.
The same goes to `untar` and `unzip`, I believe they are also are too specialized to be an extension of `File` class.

hsz: 👍 Reverted.

PR #665, opened 2021-05-18
Use Ivy instead of DownloadAction task

hsz:


src/main/kotlin/org/jetbrains/intellij/tasks/RunPluginVerifierTask.kt
```diff
@@ -336,19 +329,19 @@ open class RunPluginVerifierTask @Inject constructor(
...
-                    execute()
+                val repository = repositoryHandler.ivy { ivy ->
+                    ivy.url = URI(url)
+                    ivy.patternLayout { it.artifact("") }
+                    ivy.metadataSources { it.artifact() }
                 }
+                val dependency = dependencyHandler.create("com.jetbrains:ides:$type-$version-$buildType@tar.gz")
 
                 try {
+                    val ideArchive = configurationContainer.detachedConfiguration(dependency).singleFile
```
vlsi, line 343: Unfortunately, `detachedConfiguration` have weird names in the UI (see https://github.com/gradle/gradle/issues/17095), so you might be better creating a named configuration instead, so the user knows better what they download

hsz: That's a good point, fixed. Thanks for the review!

src/main/kotlin/org/jetbrains/intellij/dependency/CustomPluginsRepository.kt
```diff
@@ -66,22 +63,22 @@ class CustomPluginsRepository(repositoryUrl: String) : PluginsRepository {
...
-        // todo: a better way to define cache directory
-        val gradleHomePath = project.gradle.gradleUserHomeDir.absolutePath
-        val mavenCacheDirectoryPath = Paths.get(gradleHomePath, "caches/modules-2/files-2.1").toString()
-        val digest = MessageDigest.getInstance("SHA-1").digest(repositoryUrl.toByteArray())
-        val hash = Hex.encodeHex(digest).toString()
-        return Paths.get(mavenCacheDirectoryPath, "com.jetbrains.intellij.idea", hash).toString()
-    }
+    private fun downloadZipArtifact(project: Project, url: String, plugin: PluginDependencyNotation): File? {
+        val repository = project.repositories.ivy { ivy ->
+            ivy.url = URI(url)
```
vlsi, line 68: Just wondering: is the repository declared only once or does it get declared on every execution?

PR #670, opened 2021-05-25
Gha test gradle version

hsz:


zolotov:
LGTM. Please add message to all `assumeFalse` calls.

PR #679, opened 2021-05-27
Fix FAQ answer to support lazy configuration

gharrma:
With gradle-intellij-plugin 1.0, the value of `intellij.pluginName.toString()` is:
```
task ':prepareSandbox' property 'pluginName'
```

Thus we need to use `intellij.pluginName.get()` instead.


hsz:
Thanks, Matthew!

PR #697, opened 2021-06-13
Fixes #696

### LINKED ISSUE #696: buildSearchableOptions fails on MacOS
Running the `buildSearchableOptions` task on MacOS fails with the following error message on 1.0

```
Execution failed for task ':myProject:buildSearchableOptions'.
> Failed to query the value of task ':myProject:buildSearchableOptions' property 'projectExecutable'.
   > Array contains no element matching the predicate.
```

It fails here:
https://github.com/JetBrains/gradle-intellij-plugin/blob/55443b8651ad436eff70d6b3907cfd24f57e4f9c/src/main/kotlin/org/jetbrains/intellij/jbr/JbrResolver.kt#L101

AFAICS this should be
```kt
        val jbr = javaHome.listFiles()?.firstOrNull { it.name == "jbr" || it.name == "jbrsdk" }
```
instead as the following code explicitly handles the `null` case for MacOS.

DreierF:
Please see #696 for details

### DIFF
--- a/src/main/kotlin/org/jetbrains/intellij/jbr/JbrResolver.kt
+++ b/src/main/kotlin/org/jetbrains/intellij/jbr/JbrResolver.kt
@@ -98,7 +98,7 @@ open class JbrResolver @Inject constructor(
     }
 
     private fun getJbrRoot(javaHome: File): File {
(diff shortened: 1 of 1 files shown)


zolotov:
Thanks!

PR #699, opened 2021-06-15
Provide source artifacts distribution of PY/PC for building plugins

sproshev:
Update gradle-intellij-plugin to support sources for plugins for PyCharm Community and Professional Editions.
Don't hesitate to correct my code style :)

### DIFF
--- a/src/main/kotlin/org/jetbrains/intellij/dependency/PluginDependencyManager.kt
+++ b/src/main/kotlin/org/jetbrains/intellij/dependency/PluginDependencyManager.kt
@@ -99,7 +99,8 @@ open class PluginDependencyManager @Inject constructor(
     private fun registerRepositoryIfNeeded(project: Project, plugin: PluginDependency) {
         if (ivyArtifactRepository == null) {
             ivyArtifactRepository = project.repositories.ivy {
-                it.ivyPattern("$cacheDirectoryPath/[organisation]/[module]-[revision].[ext]") // ivy xml
+                val ivyFileSuffix = plugin.getFqn().substring("${plugin.id}-${plugin.version}".length)
+                it.ivyPattern("$cacheDirectoryPath/[organisation]/[module]-[revision]$ivyFileSuffix.[ext]") // ivy xml
                 it.artifactPattern("${ideaDependency?.classes}/plugins/[module]/[artifact](.[ext])") // builtin plugins
                 it.artifactPattern("$cacheDirectoryPath(/[classifier])/[module]-[revision]/[artifact](.[ext])") // external zip plugins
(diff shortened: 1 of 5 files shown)


zolotov:
lgtm. does it work? :)

sproshev:
Sources are attached to IDE:


Sources are attached to plugin:

src/main/kotlin/org/jetbrains/intellij/dependency/PluginDependencyManager.kt
```diff
@@ -106,7 +99,8 @@ open class PluginDependencyManager @Inject constructor(
     private fun registerRepositoryIfNeeded(project: Project, plugin: PluginDependency) {
         if (ivyArtifactRepository == null) {
             ivyArtifactRepository = project.repositories.ivy {
-                it.ivyPattern("$cacheDirectoryPath/[organisation]/[module]-[revision].[ext]") // ivy xml
+                val ivyFileSuffix = plugin.getFqn().substring("${plugin.id}-${plugin.version}".length)
```
zolotov, line 102: I think it won't work. for `plugin-123-2.xml` prefix would be equal to `-2.xml` and according to pattern the following file will be expected: `$cacheDirectoryPath/[organisation]/[module]-[revision]-2.xml.xml`

src/main/kotlin/org/jetbrains/intellij/dependency/PluginDependencyManager.kt
```diff
@@ -148,7 +142,8 @@ open class PluginDependencyManager @Inject constructor(
                 addArtifact(IntellijIvyArtifact.createDirectoryDependency(it, configuration.name, baseDir, groupId))
             }
             ideaDependency?.sources?.takeIf { plugin.builtin }?.let {
-                val artifact = IntellijIvyArtifact(it, "ideaIC", "jar", "sources", "sources")
+                val name = if (isDependencyOnPyCharm(ideaDependency)) "pycharmPC" else "ideaIC"
```
zolotov, line 145: I think it's ok, but for even better experience it makes sense to download pycharm sources and attach them to Python plugin dependency even if `ideaDependency` is IntelliJ

zolotov (changes requested):
please fix suffix, see IdeaDependencyManager#register

src/main/kotlin/org/jetbrains/intellij/dependency/PluginDependencyManager.kt
sproshev: `plugin.getFqn()` does not have `.xml` as a suffix:
pluginFqn

zolotov (approved):

PR #702, opened 2021-06-18
Configure classpath after running executable

awildbrysen:
While running older versions of IntelliJ (f.e. 2018.2) the `StartupUtil#checkJdkVersion` call blocks the launch because of the runtime being used missing the `com.sun.jdi.Field` class, which is present in the tools.jar which is being resolved by `RunIdeBase#resolveToolsJar`, but in the current state this seems to be unreachable code since executable is always null


hsz:
Nice catch, @awildbrysen – thank you!

PR #705, opened 2021-06-18
Fix cacheability of the test task

cesar1000:
Fix several issues that affect the cacheability of the test task:

* System properties containing absolute paths should be set in `doFirst`, so that they are not captured in the cache key (which would make the cache entry non-relocatable).
* The configuration directory should be declared as an input, since it's created outside of the task, and Gradle will disable caching upon finding unexpected changes to an output directory.
* All file inputs should use relative path sensitivity in order to make the task relocatable.
* Passing a lambda into `doFirst` disables caching. Gradle recommends using an anonymous class instead.
JIRA Issues: ANDROID-76125


hsz:
Great job, Cesar – thanks!

cesar1000:
So fast! Thanks :))

PR #707, opened 2021-06-20
Use classpath normalization for the plugin directory input in tests

cesar1000:
The plugin directory needs to be normalized as a classpath so that class timestamps and class order within the jar files are not considered when evaluating if the task is up to date. A regular binary comparison typically breaks the relocation of the cached artifacts.

Also, I added a comment explaining a previous change, in order to warn future maintaines about reverting it accidentally.


hsz:
Thanks for your contribution, Cesar!

PR #742, opened 2021-07-14
Move signPlugin file name creation to lazy

brian-mcnamara:
Related to: https://jetbrains-platform.slack.com/archives/CPL5291JP/p1626141308142300

Fixes issue where plugin version is specified in the build.gradle file where the filename with the version is generated after initialization. This prevented the output of signPlugin from containing the version.


zolotov:
thanks, it looks very good to me. @hsz please take a look too

hsz:
Thanks, Brian!

PR #744, opened 2021-07-15
Add support for GW

abrooksv:


src/main/kotlin/org/jetbrains/intellij/utils.kt
```diff
@@ -201,7 +202,8 @@ fun releaseType(version: String) = when {
     version.endsWith("-EAP-SNAPSHOT") ||
         version.endsWith("-EAP-CANDIDATE-SNAPSHOT") ||
         version.endsWith("-CUSTOM-SNAPSHOT") ||
-        version.matches(MAJOR_VERSION_PATTERN.toRegex())
+        version.matches(MAJOR_VERSION_PATTERN.toRegex()) ||
+        BUILD_VERSION_PATTERN.matcher(version).matches()
```
zolotov, line 206: this one is not good. I think it breaks nightly IDEA builds, that look like `201.1463-SNAPSHOT` and must be downloaded from `nightly` repo. 
GW should be either uploaded to `nightly` (access from JB network only), or to be renamed, e.g. with `-CUSTOM-SNAPSHOT` prefix (I cannot remember why we introduced it, though).

zolotov (changes requested):

shalupov:
I merged maven coordinates as https://github.com/JetBrains/gradle-intellij-plugin/commit/5f078a51d433cad10e37331ae5c49df76af74682

As for other changes we discussed with @zolotov how not to do them at all.
So, support for JetBrains Gateway and any future product will be added around next week

zolotov:
see https://github.com/JetBrains/gradle-intellij-plugin/pull/746. Going to include that in 1.1.4

PR #746, opened 2021-07-19
Platform prefix

zolotov:


hsz (approved):

src/main/kotlin/org/jetbrains/intellij/tasks/RunIdeBase.kt
```diff
@@ -145,18 +124,22 @@ abstract class RunIdeBase(runAlways: Boolean) : JavaExec() {
...
+                    .asSequence()
+                    .filter { file -> file.fileName.toString().endsWith(".sh") }
+                    .flatMap { file -> Files.lines(file).asSequence() }
+                    .mapNotNull { line -> platformPrefixSystemPropertyRegex.find(line)?.groupValues?.getOrNull(1) }
+                    .firstOrNull()
+
+            if (prefix != null) {
+                systemProperty("idea.platform.prefix", prefix)
             }
+            info(context, "Using idea.platform.prefix=$prefix")
```
shalupov, line 138: IMHO you need to assert the case where prefix was not found by both system property and script search

zolotov: > where prefix was not found by both system property and script search

Unfortunately, it means that empty prefix should be used for IntelliJ Ultimate

shalupov: Yes, in that case, you may check that it's really IntelliJ Ultimate and not some other IDE

zolotov: makes sense. done

Issue #913, opened 2022-03-02
Add additional customization to publishPlugin task


KonstantinHudyakov:
I faced with a problem of uploading the plugin to Toolbox Enterprise repository. It is not possible due to some configuration limitations of `publishPlugin` task.
1. I can specify the host, but can not specify the route for uploading request. For now it always sends the request to `<host>/plugin/uploadPlugin`.
2. Toolbox Enterprise requires `Automation` authorization token. But this task sends the request only with `Bearer` token.

I think that additional parameters like `route` and `authorizationType` can be added to configuration of `publishPlugin` task.

PR #919, opened 2022-03-17
Initial Toolbox Enterprise integration #913

hsz:


KonstantinHudyakov (commented):
Also the convention for `toolboxEnterprise` property should be specified in the `IntelliJPlugin.configurePublishPluginTask()` because for now the exception will be thrown if the property is not set directly.

PR #948, opened 2022-04-12
Check Kotlin version

hsz:
Check if used Kotlin version matches the bundled one.

TODO:
- [ ] cover with integration tests
- [ ] extend docs with explanation about `compileKotlin.kotlinOptions.apiVersion`
- [ ] make the IDE-Kotlin versions map from [docs](https://plugins.jetbrains.com/docs/intellij/kotlin.html#kotlin-standard-library) available as a JSON list – In the intellij-sdk-docs repository
- [ ] fetch above list automatically in the Gradle IntelliJ Plugin

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## next
 - Fixed code instrumentation disabling via `tasks.instrumentCode.enabled`
+- Check if used Kotlin version matches the bundled one
 
 ## 1.5.1
 - Add `util_rt.jar` to the classpath of run-like tasks for `2022.1+` compatibility
--- a/src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
+++ b/src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
@@ -48,8 +48,11 @@ import org.jetbrains.intellij.pluginRepository.PluginRepositoryFactory
 import org.jetbrains.intellij.tasks.*
 import org.jetbrains.intellij.utils.ArchiveUtils
 import org.jetbrains.intellij.utils.DependenciesDownloader
+import org.jetbrains.intellij.utils.getBundledKotlinStdlib
 import org.jetbrains.intellij.utils.ivyRepository
 import org.jetbrains.intellij.utils.mavenRepository
+import org.jetbrains.kotlin.gradle.plugin.getKotlinPluginVersion
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.io.File
 import java.net.URL
 import java.time.LocalDateTime
@@ -72,6 +75,7 @@ open class IntelliJPlugin : Plugin<Project> {
         context = project.logCategory()
 
         checkGradleVersion(project)
+        checkKotlinVersion(project)
         project.plugins.apply(JavaPlugin::class.java)
         project.plugins.apply(IdeaExtPlugin::class.java)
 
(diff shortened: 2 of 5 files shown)


CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## next
 - Fixed code instrumentation disabling via `tasks.instrumentCode.enabled`
+- Check if used Kotlin version matches the bundled one
```
YannCebron, line 5: AFAIU it's not only Kotlin version, but also apiCompatibility/stdlib issue. please clarify (see also notes about output messages)

> "...matches the one bundled in target IDE"

src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
```diff
@@ -116,6 +120,54 @@ open class IntelliJPlugin : Plugin<Project> {
...
+                println("kotlinLanguageVersion='${kotlinLanguageVersion}'")
+                println("kotlinApiVersion='${kotlinApiVersion}'")
+
+                val stdlib = "Kotlin Standard Library (kotlin-stdlib)"
+                val ide = "IntelliJ Platform ($ideVersion)"
+
+                when (kotlinBundled) {
+                    true -> when (ideKotlinVersion) {
+                        null -> null // This is fine. Kotlin not bundled in the given version as: Since Kotlin 1.4, a dependency on the standard library stdlib is added automatically.
+                        else -> "You bundle $stdlib with your plugin. Please consider removing it to avoid potential issues in the runtime and decrease the size of your archive."
```
YannCebron, line 151: remove "You". clarify "potential issues" (link?)

src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
```diff
@@ -116,6 +120,54 @@ open class IntelliJPlugin : Plugin<Project> {
...
+                    }
+
+                    false -> when {
+                        ideKotlinVersion == null -> "$stdlib is not bundled in the $ide."
```
YannCebron, line 155: I assume we cannot access exact configuration of `intellij` here? so we could print out "target platform XYZ in version Z.Y." or similar?

src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
```diff
@@ -116,6 +120,54 @@ open class IntelliJPlugin : Plugin<Project> {
...
+                        kotlinVersion == ideKotlinVersion -> null // Kotlin versions match, we're good
+                        kotlinVersion < ideKotlinVersion -> "$stdlib version is lower than the one bundled within the $ide, consider upgrading your Kotlin plugin to: $ideKotlinVersion"
```
YannCebron, line 157: clarify what "upgrading" means exactly.
clarify "Kotlin plugin" refers to the one defined in Gradle build script. not eventually also existing dependency on Kotlin plugin for the project plugin.

src/main/kotlin/org/jetbrains/intellij/utils/BundledKotlinStdlib.kt
```diff
@@ -0,0 +1,33 @@
+package org.jetbrains.intellij.utils
+
+import org.jetbrains.intellij.Version
+
+private val bundledStdlibApiMap = mapOf(
```
YannCebron, line 5: we can't do that. bugfix releases could change it. also what about AS or other 3rd party IDEs?

can't we scan $IDE$/lib/kotlin-stdlib.jar?

src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
```diff
@@ -116,6 +120,54 @@ open class IntelliJPlugin : Plugin<Project> {
         }
     }
 
+    private fun checkKotlinVersion(project: Project) {
+        project.afterEvaluate {
+            // Current project is Kotlin-based, as Kotlin JVM plugin is loaded
```
YannCebron, line 125: WDYM with "Kotlin-based"? Kotlin JVM plugin is defined in Gradle setup?

src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
```diff
@@ -116,6 +120,54 @@ open class IntelliJPlugin : Plugin<Project> {
...
+                val ideVersion = setupDependenciesTask.idea.get().version
+                val ideKotlinVersion = getBundledKotlinStdlib(ideVersion)?.let(Version.Companion::parse)
+                val ideKotlinVersionNoPatch = ideKotlinVersion?.run { Version(major, minor, 0) } ?: Version()
+                val kotlinVersion = project.getKotlinPluginVersion().let(Version.Companion::parse)
+                val kotlinVersionNoPatch = Version(kotlinVersion.major, kotlinVersion.minor, 0)
+                val kotlinApiVersion = kotlinCompileTask.kotlinOptions.apiVersion?.let(Version.Companion::parse)
+                val kotlinLanguageVersion = kotlinCompileTask.kotlinOptions.languageVersion?.let(Version.Companion::parse)
+                val kotlinBundled = project.property("kotlin.stdlib.default.dependency").toString().toBoolean()
+
+                println("kotlinLanguageVersion='${kotlinLanguageVersion}'")
```
YannCebron, line 142: remove debug

YannCebron (changes requested):

YannCebron:
I'd rather scan the Kotlin stdlib from the actual target platform bundled library binary. it's safer, will automatically support newer target IDE versions, and also requires no internet.
also this probably needs buildFeature flag

Issue #786, opened 2021-09-20
Automatically detect bundled sources in plugin dependency


artspb:
The Go plugin has a separate `go-openapi.jar` and ships it together with its sources. The plugin now has the following structure:

```
go
  | go-openapi.jar
  | src
    | go-openapi-src.jar
```

I know that some other plugins have the same structure. It'd be great if we can we make gradle-intellij-plugin attach sources automatically.

zolotov:
relates to #207

artspb:
Oh, nice. Might be even a duplicate.

YannCebron:
I like this approach, as it would allow us to provide "open api" sources for _any_ plugin used as dependency, including 3rd party. Without need of Marketplace to provide source artifacts separately.

Issue #153, opened 2016-11-24
Sources of bundled plugins are not attached in IDE


zolotov:
```
intellij { plugins 'coverage' }
```

Expected that coverage sources will be attached to the project

zolotov:
\cc @da1z

PR #954, opened 2022-04-14
Re: #786: Automatically detect bundled sources in plugin dependency 

karollewandowski:
Fixes: #786.

In the case when multiple source artifacts exist for a plugin, only the first one will be attached because of the IDEA issue: [IDEA-292168](https://youtrack.jetbrains.com/issue/IDEA-292168).

This may be the case when bundled plugin contains sources - IDE sources are attached for bundled plugins so we will have plugin and IDE sources attached (attaching IDE sources was implemented in the scope of #153). I intentionally ordered plugin sources first to avoid [IDEA-292168](https://youtrack.jetbrains.com/issue/IDEA-292168).

I was wondering whether we should configure sources when `intellij.downloadSources` is false and plugin is bundled, but I think it's better to add them always whenever we have them without downloading. What may look strange is that Ivy file will contain plugin's source entry in "regular" and `-withSources` versions. Let me know if I should change it.

Additional small cleanups:
- added default predicate to `collectJars()` for cleaner client code
- removed `PluginDependency.sourcesDirectory` as it is legacy and not used at all
- added clarification about including gradle plugin project in test IntelliJ plugin project for debugging purposes


karollewandowski:
Windows runners tests failed due to: `java.io.IOException: There is not enough space on the disk`.

karollewandowski:
> Windows runners tests failed due to: `java.io.IOException: There is not enough space on the disk`. Is it known/addressed?

Windows runners have 14 GB of space on disk D which is used for executing actions (see: https://github.com/actions/virtual-environments/issues/1341). We've used this limit after my changes because I added tests requiring the GoLand, and IDEA 2021.1.4 (additional 1.46 GB + 2.3 GB = 3.76 GB).
2021.1.4 is the lowest version compatible with the oldest Go plugin having `/lib/src` artifacts (I've checked PHP and Python plugins which do have sources as well, but they don't have them in versions older than 2021.1.4).

As the solution, I will update our default IDEA version used in tests to 2021.1.4 and will adjust other tests where needed.

This issue may happen in the future again if we download more files/IDEs, but the alternative solution is not trivial (see https://github.com/actions/virtual-environments/issues/1341) and maybe we won't need it for a longer time.

karollewandowski:
> > Windows runners tests failed due to: `java.io.IOException: There is not enough space on the disk`. Is it known/addressed?
> 
> Windows runners have 14 GB of space on disk D which is used for executing actions (see: [actions/virtual-environments#1341](https://github.com/actions/virtual-environments/issues/1341)). We've used this limit after my changes because I added tests requiring the GoLand, and IDEA 2021.2.4 (additional 1.46 GB + 2.3 GB = 3.76 GB). 2021.2.4 is the lowest version compatible with the oldest Go plugin having `/lib/src` artifacts. I've checked PHP and Python plugins which do have sources as well for a potential alternative compatible with currently used 2020.1, but they don't have them in versions older than 2021.2.4.
> 
> As the solution, I will update our default IDEA version used in tests to 2021.2.4 and will adjust other tests where needed.
> 
> This issue may happen in the future again if we download more files/IDEs, but the alternative solution is not trivial (see [actions/virtual-environments#1341](https://github.com/actions/virtual-environments/issues/1341)) and maybe we won't need it for a longer time.

I updated the default tested platform version to 2021.2.4 and we have ~2 GB margin now.
We can decrease it by 0.5 GB by getting rid of IC-14.1.4 usage in some tests, but I'm leaving it for now.

src/test/kotlin/org/jetbrains/intellij/tasks/DownloadIntelliJSpec.kt
```diff
@@ -5,100 +5,109 @@ import org.jetbrains.intellij.IntelliJPluginSpecBase
 import java.io.File
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 
 class DownloadIntelliJSpec : IntelliJPluginSpecBase() {
```
karollewandowski, line 11: If I understood these tests correctly, they were fragile because files could exist because other tests, executed before, downloaded them. I'm making sure all relevant stuff is deleted in the setup method. Tests are slower, but safe now.

Also:
- refactored tests for better readability and added one case so we have now downloading ultimate IDE with and without sources
- changed IU 14.1.1 to 2021.2.4 to save some disk space on the runners

.github/workflows/reusable-unitTests.yml
```diff
@@ -31,7 +31,8 @@ jobs:
           cache: gradle
 
       - name: Run Tests
-        run: ./gradlew check -PtestGradleVersion="${{ matrix.gradleVersion }}"
+        # cache test Gradle instance files on disk C: on Windows due to the limited space on D:
+        run: ./gradlew check -PtestGradleVersion="${{ matrix.gradleVersion }}" ${{ runner.os == 'Windows' && '-PtestGradleUserHome="C:\\testGradleHome"' || '' }}
```
karollewandowski, line 35: Because disk space on disk D is limited.

src/test/kotlin/org/jetbrains/intellij/IntelliJPluginSpec.kt
```diff
@@ -1,14 +1,13 @@
 package org.jetbrains.intellij
 
 import org.gradle.api.plugins.JavaPlugin
+import org.gradle.testkit.runner.BuildResult
```
karollewandowski, line 4: [Applies to entire file]
I cleaned the tests:
- reduced code duplication
- reduces code noise in testing methods, so they include actual tested context and assertions
- added messages to assertions - before there was a lot of "expected true, but got false" messages and now there will be more precise

All existing tests and logic should be preserved.

src/test/kotlin/org/jetbrains/intellij/tasks/DownloadIntelliJSpec.kt
karollewandowski: This is updated now to use unique IDEs versions. Unfortunately, deleting directories before tests fail on Windows runners, probably due to the daemon locking the files, which shouldn't be run at all in our setup. It may be caused by Gradle issues, see:
- https://github.com/gradle/gradle/issues/937
- https://github.com/gradle/gradle/issues/12535

I'm leaving this approach for now, which is better than nothing, and will monitor the above issues. We can improve it in the future.

src/test/kotlin/org/jetbrains/intellij/IntelliJPluginSpec.kt
```diff
@@ -252,23 +161,147 @@ class IntelliJPluginSpec : IntelliJPluginSpecBase() {
...
+
+        assertContainsOnlySourceArtifacts(result,
+            "go/lib/src/go-openapi-src-212.5712.14-unzipped.com.jetbrains.plugins.jar " +
+                    "(unzipped.com.jetbrains.plugins:org.jetbrains.plugins.go:212.5712.14)"
+        )
+    }
+
+    @Test
+    fun `add bundled zip plugin source artifacts from src directory when localPath used`() {
+        val localPath = createLocalIdeIfNotExists(
```
karollewandowski, line 250: This downloads and unpacks IDEs next to the test Gradle home directory.
I created custom logic for downloading IDEs to not depend on the plugin logic (if reusing is even possible). It can be also used in other tests requiring local installations.

src/test/kotlin/org/jetbrains/intellij/tasks/DownloadIntelliJSpec.kt
```diff
@@ -11,90 +11,124 @@ import kotlin.test.assertTrue
...
-    private val iuDistCacheDir = File(iuCacheDir, "4493a5312869afef74b9cc9f0223b40794762ea4")
-
-    override fun setUp() {
-        super.setUp()
-        // delete only dist and sources directories, as POM dir is not recreated
-        deleteIfExists(icSourcesCacheDir)
-        deleteIfExists(icDistCacheDir)
-        deleteIfExists(iuDistCacheDir)
-    }
+    private val icCacheDir = File(gradleHome, "caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC")
```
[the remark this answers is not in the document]
karollewandowski: I understand you mean `modules-2` and `files-2.1`. They are the same in all the versions we support (on all operating systems) and they seem to be stable. I've found information that they are related to the files structure layout and `modules-2` is present from Gradle 1.9. They are frequently referenced in Gradle docs literally so I wouldn't worry they will change soon.

src/test/kotlin/org/jetbrains/intellij/test/localIdeTestUtils.kt
```diff
@@ -0,0 +1,56 @@
...
+    val fileName = releasePath.substringAfterLast('/')
+    val localIdeZipPath = localIdesPath.resolve(fileName)
+    val localIdeDirPathString = localIdeZipPath.toString().removeSuffix(".zip")
+    if (Files.exists(Path.of(localIdeDirPathString))) {
+        return localIdeDirPathString
+    }
+    if (!Files.exists(localIdesPath)) {
+        Files.createDirectories(localIdesPath)
+    }
+    URL("${IntelliJPluginConstants.DEFAULT_INTELLIJ_REPOSITORY}/releases/$releasePath")
```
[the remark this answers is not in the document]
karollewandowski: Unfortunately, I don't see anything easily usable out of Gradle context.

src/test/kotlin/org/jetbrains/intellij/test/localIdeTestUtils.kt
```diff
@@ -0,0 +1,56 @@
...
+        .openStream().use { Files.copy(it, localIdeZipPath) }
+    localIdeZipPath.toFile().unzip()
+    Files.delete(localIdeZipPath)
+    return localIdeDirPathString
+}
+
+private fun File.unzip() {
+    val unzipDirPath = this.path.removeSuffix(".zip")
+    ZipFile(this).use { zip ->
```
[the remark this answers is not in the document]
karollewandowski: I don't see anything easily usable out of Gradle context. I considered using Zip4j but wanted to avoid adding a dependency. I'm not sure what is less evil here (dependency vs 15 lines of code).
Let me know if I should improve it.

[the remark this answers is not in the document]
karollewandowski: Agreed to skip it for now.

karollewandowski:
Agreed with Yann to skip a feature flag for this as it would introduce unnecessary complexity.
Attaching sources depends on generating Ivy files and if we wanted to isolate the old logic of attaching IDE sources and control only attaching bundled sources, we would have to generate unique Ivy files for each context combination (no sources, IDE sources only, bundled sources only, IDE+bundled sources) and it would grow if we added new ways of attaching sources. An alternative is to control attaching sources as a whole, but we decided to skip this as it is a pretty safe feature and doesn't affect building the project itself.

I'm delaying the merge until the new version is released. I'll rebase the branch and fix failed checks after the release.

YannCebron (approved):

PR #969, opened 2022-04-26
Build Features

hsz:


src/main/kotlin/org/jetbrains/intellij/utils.kt
```diff
@@ -219,3 +221,18 @@ fun Boolean.ifFalse(block: () -> Unit): Boolean {
...
 }
+
+fun Boolean.ifTrue(block: () -> Unit): Boolean {
+    if (this) {
+        block()
+    }
+    return this
+}
+
+internal fun Project.isEnabled(feature: String) = (project.findProperty(feature) == "true").ifTrue {
```
YannCebron: - `isBuildFeatureEnabled` would be more clear IMHO
- why not enum as arg?

src/main/kotlin/org/jetbrains/intellij/IntelliJPluginConstants.kt
```diff
@@ -43,6 +43,9 @@ object IntelliJPluginConstants {
     const val RELEASE_TYPE_NIGHTLY = "nightly"
     const val RELEASE_TYPE_RELEASES = "releases"
 
+    const val BUILD_FEATURE_CHECK_GRADLE_VERSION = "org.jetbrains.intellij.buildFeature.checkGradleVersion"
+    const val BUILD_FEATURE_CHECK_GRADLE_INTELLIJ_PLUGIN_VERSION = "org.jetbrains.intellij.buildFeature.checkGradleIntellijPluginVersion"
```
YannCebron, line 47: why do we re-define the property name here? re-use enum name or add `configPropertyName` property to enum and externalize prefix

src/main/kotlin/org/jetbrains/intellij/utils.kt
```diff
@@ -219,3 +221,18 @@ fun Boolean.ifFalse(block: () -> Unit): Boolean {
...
 }
+
+fun Boolean.ifTrue(block: () -> Unit): Boolean {
+    if (this) {
+        block()
+    }
+    return this
+}
+
+internal fun Project.isBuildFeatureEnabled(feature: String) = (project.findProperty(feature) == "true").ifTrue {
```
YannCebron, line 232: use `BuildFeature` as parameter type

src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
```diff
@@ -113,22 +116,32 @@ open class IntelliJPlugin : Plugin<Project> {
...
+        if (project.isBuildFeatureDisabled(BUILD_FEATURE_CHECK_GRADLE_VERSION)) {
+            return
+        }
         if (Version.parse(project.gradle.gradleVersion) < Version.parse("6.7")) {
             throw PluginInstantiationException("${IntelliJPluginConstants.NAME} requires Gradle 6.7 and higher")
         }
     }
 
     private fun checkPluginVersion(project: Project) {
+        if (project.isBuildFeatureDisabled(BUILD_FEATURE_CHECK_GRADLE_INTELLIJ_PLUGIN_VERSION)) {
```
YannCebron, line 128: what about `CI`? automatically disable this check? all checks? or ability to override `CI` check default via property in `BuildFeature` enum?

src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
```diff
@@ -113,22 +116,32 @@ open class IntelliJPlugin : Plugin<Project> {
...
+            return
```
YannCebron, line 129: IMHO it would be safer to disable this check by default (for now). needs more research about caveats

integration-tests/build-features/verify.main.kts
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env -S kotlin -J-ea
+
+@file:Import("../verify.utils.kts")
+
+with(__FILE__.toPath()) {
+    runGradleTask(mapOf(
+        "org.jetbrains.intellij.buildFeature.checkGradleVersion" to false,
```
YannCebron, line 7: same here, String vs enum

src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
hsz: Why shouldn't that be checked on CI? If someone pushes changes to the remote without checking it locally, that could be the only way to react to an unexpected state.

src/main/kotlin/org/jetbrains/intellij/IntelliJPluginConstants.kt
hsz: Well, enum was a leftover. But constructing consts like that – I don't like it anymore and I'll get back to enums now. 🤷‍♂️

src/main/kotlin/org/jetbrains/intellij/IntelliJPluginConstants.kt
```diff
@@ -43,6 +43,9 @@ object IntelliJPluginConstants {
     const val RELEASE_TYPE_NIGHTLY = "nightly"
     const val RELEASE_TYPE_RELEASES = "releases"
 
+    const val BUILD_FEATURE_CHECK_GRADLE_VERSION = "org.jetbrains.intellij.buildFeature.checkGradleVersion"
```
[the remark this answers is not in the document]
hsz: This check requires an internet connection which some environments lack.

src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
YannCebron: By default, we should assume CI has no network access.

Issue #207, opened 2017-06-26
Attach `<IDE_ROOT>/lib/src/*.zip` sources


zdary:
Hi,

I have a question. While using gradle plugin for plugin development I chose to download sources as well. The sources are downloaded but When I navigate to a class from my plugin to a class in open-api I always end up in decompiled class instead of source code downloaded by gradle.

IDEA comes with src forlder for spring.api and a few more. Is there a way gradle pluging would attach those as sources?

zolotov:
IntelliJ Utlimate is not open sources project. You can open only sources of classes from IntelliJ Community.

> IDEA comes with src forlder for spring.api and a few more

I don't see any src directory for spring. Spring is a close-sourced part.

zolotov:
closing due to lack of feedback. the issue is unclear to me

bjansen:
@zolotov the Ultimate edition comes with a few OpenAPI extensions for Ultimate plugins, in order to allow developers to contribute extensions points to these plugins I suppose. Spring is one of them, the source code for Spring's OpenAPI is located under `$IDEA_HOME$/lib/src/src_spring-openapi.zip`. This folder also contains source code for other plugins like Tomcat, persistence etc. Perhaps the gradle-intellij-plugin could automatically attach these source files to their corresponding plugins?

See this page for more info: http://www.jetbrains.org/intellij/sdk/docs/reference_guide/frameworks_and_external_apis/spring_api.html

zolotov:
@bjansen found it, thank you!

PR #1019, opened 2022-06-13
Attaching plugin source from <IDE_ROOT>/lib/src directory (#207)

karollewandowski:
# Pull Request Details

Attaches plugin source from `<IDE_ROOT>/lib/src` directory.

## Description

Sources are provided only with IntelliJ IDEA Ultimate distribution. I've checked IntelliJ IDEA Community, RubyMine, WebStorm, and PhpStorm - none of them contains the `lib/src` directory.

Technical solution:
- source ZIPs are collected as part of the IDE dependency object
- `PluginDependencyManager` adds source entries in the plugin's Ivy file
- a new Ivy artifact pattern is registered: `<IDE_ROOT>/[artifact](.[ext])` to be able to resolve source artefacts specified as e.g.:
```xml
<artifact name="lib/src/src_css-api" type="zip" ext="zip" conf="sources"/>
```

The implementation uses explicit mapping of the platform plugin ID to the source's file name. This is needed because in the IDE distribution there is no information linking source ZIPs with specific plugins (file names are not enough). The implementation will be removed in the future as we will aim to provide sources within plugin ZIP distributions, not in the IDE.

If the plugin bundle contains source JAR, it will take precedence over the sources in IDE distribution (it shouldn't happen).

I've added the `platformPluginId` property in `PluginDependency` for resolving a source ZIP file defined in the mapping. We could potentially get rid of it and use IDs used internally in the Gradle repository, but I think it can become unstable when plugin directory names (e.g. `IDE_ROOT/plugins/SpringBoot`) are changed.

## Related Issue

#207: Attach `<IDE_ROOT>/lib/src/*.zip` sources

## Motivation and Context

It allows developers to use open API sources while developing plugins, e.g., when debugging issues.

## How Has This Been Tested

Tested locally with the test plugin project. Build script included:
```kotlin
intellij {
    type.set("IU")
    version.set("2021.2.4")
    plugins.set(listOf(
        "com.intellij.css",
        "com.intellij.database",
        "com.intellij.javaee",
        "com.intellij.spring",
        "com.intellij.spring.boot",
        "com.intellij.persistence",
        "Tomcat"
    ))
}
```

The Gradle IntelliJ Plugin generated Ivy files with source ZIP artifacts entries. After importing the project in IDE, sources were available when some plugin class was open e.g. `com.intellij.psi.css.CssElementDescriptorProvider`).

## Types of changes

## Checklist

- [ ] My change requires a change to the [documentation](https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html).
- [x] I have included my change in the [**CHANGELOG**](https://github.com/JetBrains/gradle-intellij-plugin/blob/master/CHANGES.md).

### DIFF
--- a/.github/scripts/list_integration_test_modules.main.kts
+++ b/.github/scripts/list_integration_test_modules.main.kts
@@ -13,14 +13,12 @@ import kotlin.streams.toList
 val rootPath: Path = System.getenv("GITHUB_WORKSPACE")?.let(Path::of)
     ?: Path.of("").toAbsolutePath().parent.parent
 
-// FIXME: revert before merging
-/*val modules: List<String> = rootPath
+val modules: List<String> = rootPath
     .let { it.resolve("integration-tests") }
     .let { Files.list(it) }
     .filter { Files.isDirectory(it) }
     .map { it.fileName.toString() }
     .filter { !it.startsWith(".") }
-    .toList()*/
-val modules: List<String> = listOf("attaching-plugin-sources-from-ide-dist")
+    .toList()
 
 println("[\"${modules.joinToString("\", \"")}\"]")
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,10 +14,10 @@ jobs:
     name: Gradle Validation
     uses: ./.github/workflows/reusable-gradleValidation.yml
 
-#  unitTests: # FIXME: revert before merging
-#    name: Unit Tests
-#    needs: gradleValidation
-#    uses: ./.github/workflows/reusable-unitTests.yml
+  unitTests:
+    name: Unit Tests
+    needs: gradleValidation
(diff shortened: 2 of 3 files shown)


YannCebron:
if sources are available from the plugin itself, they must take precedence to those bundled with the IDE distribution (just in case we have some overlap)

YannCebron (approved):

karollewandowski:
Qodana check is queued and I don't see any cancel/re-run button for this specific check (because it is queued I assume, I reported it to the Qodana team):
https://github.com/JetBrains/gradle-intellij-plugin/pull/1019/checks?check_run_id=6912165556

It has passed previously and I don't expect any violations in the code. I would like to avoid re-checking this PR as it takes more than an hour.

Please consider checks as passed.

jreznot (approved):

Issue #808, opened 2021-10-15
Merge all plugin jars into single one for new plugin model


Undin:
IntelliJ platform 2021.2 introduced new plugin model to make support for dynamic unloading easier and more stable.
Now the plugin can declare some modules inside it (slightly modified optional dependency from plugin model 1.0) with own dependencies on other plugins or modules. At the same time, the implementation of new plugin model supposes that all manifests (plugin.xml or module manifests) are located in single jar file to make classloading faster. As a result, it forces plugin writers to assemble all plugin jars into single jar file.

And it's a quite complex task for plugin writers if their plugin project consists of several Gradle modules. You can't just "hack" `jar` task of the root plugin module to create single jar because the current `PrepareSandboxTask` implementation collects all jars of plugin project into single directory that's supposed to be archived into plugin artifact and you have to come up with some workaround not to pack classes from submodules twice and at the same time not to miss jar files of external dependencies. As a result, it leads to quite weird implementations like [this one](https://github.com/JetBrains/educational-plugin/commit/562d71a53c8399bf876902dce793ae04cfd67a92) in `EduTools` plugin or [this one](https://github.com/Camelcade/Perl5-IDEA/commit/59e8aa9c8743de6c096c6acec7a60f0346dd302d) in `Perl` plugin. And strictly speaking, my implementation in `EduTools` plugin works by accident because we declare external dependencies for all Gradle modules including root one (where they are redundant) for some reason. But if we remove these "unnecessary" dependencies from the root module, the plugin artifact won't contain jars of external dependencies.

You may note that the platform doesn't force you to migrate to the new plugin model (at least, for now). But it's not strictly true because:
- dynamic unloading should work properly only with the new plugin module
- if you have a dependency on another plugin, you may face with issues like https://github.com/intellij-rust/intellij-rust/issues/7448 [^1] at any moment and, as I understand, there are only two solutions:
  * hack platform not to throw an exception in your specific case. It was done for Rust plugin to fix https://github.com/intellij-rust/intellij-rust/issues/7448 but it's not a good solution for any external plugin
  * migrate your plugin to new plugin model to provide proper dependencies
  
  
So it would be really great to simplify migration to new plugin model for all plugin writers outside of IntelliJ repository and provide proper plugin artifact creation out of the box
  
[...]

PR #1069, opened 2022-07-26
Support custom jar creation task

zielu:
# Pull Request Details

Name of jar producer task can me customized. Task is still expected to be an instance of Gradle Jar task.

## Description

By adding ability to customize name of task that produces plugin jar it is possible to either use customized jar task or some fat jar Gradle plugin (for example [Gradle Shadow](https://imperceptiblethoughts.com/shadow/)) to assemble single jar out of multiple modules.
By convention default name is set to Gradle JAR_TASK_NAME so existing plugins do not need to change anything on update.

## Related Issue

https://github.com/JetBrains/gradle-intellij-plugin/issues/808

## Motivation and Context

Support assembly of multi module plugins into single plugin jar.

## How Has This Been Tested

With this change I was able to build a single plugin jar out of project that consists of main plugin module and 5 supporting modules. 

My project set up is quite specific, let assume I have following modules:
- plugin
- module-api
- module-A
- module-B

module-A and module-B have compileOnly dependency on module-api.

plugin module defines 2 configurations:
-  apiPluginModules
-  compileOnly.extendsFrom apiPluginModules
-  runtimePluginModules

And has following dependencies:
- apiPluginModules on module-api
- runtimePluginModules on module-A and module-B

I'm using Gradle Shadow to create fat jar and it has config like:
```
shadowJar {
  archiveClassifier.set('')
  configurations = [
      project.configurations.apiPluginModules,
      project.configurations.runtimePluginModules
  ]
}
```

intellij extension is configured like this:
```
intellij {
    ...
    jarTaskName = 'shadowJar'
}
```

Plugin distribution zip was created successfully and I was able to install it and run plugin in a 2022.2 RC Idea.

## Types of changes

## Checklist

- [x] My change requires a change to the [documentation](https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html).
- [x] I have updated the documentation accordingly.
- [x] I have included my change in the [**CHANGELOG**](https://github.com/JetBrains/gradle-intellij-plugin/blob/master/CHANGES.md).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - Configure classpath for run-based tasks using `package-info.json` provided with IntelliJ SDK 2022.3+
 - The [`verifyPluginConfiguration`](https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#tasks-verifypluginconfiguration) task for validating the plugin project configuration.
 - Make Android Studio (`AI` type) available for resolving as SDK.
+- Add `jarTaskName` to `intellij` extension to allow customization of jar creation [#808](../../issues/808)
 
 ### Changed
 - Change `IntelliJPluginConstants.ANDROID_STUDIO_PRODUCTS_RELEASES_URL` to `https://jb.gg/android-studio-releases-list.xml`
@@ -566,4 +567,3 @@
 
 ## 0.0.10
 - Support for attaching IntelliJ sources in IDEA
-
--- a/src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
+++ b/src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt
@@ -130,6 +130,7 @@ open class IntelliJPlugin : Plugin<Project> {
             downloadSources.convention(!System.getenv().containsKey("CI"))
             configureDefaultDependencies.convention(true)
             type.convention(PLATFORM_TYPE_INTELLIJ_COMMUNITY)
+            jarTaskName.convention(JavaPlugin.JAR_TASK_NAME)
         }
 
         configureTasks(project, intellijExtension)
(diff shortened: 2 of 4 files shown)


zielu:
I added one more flag - this time prepareSandbox.includeRuntimeConfigurationLibraries. Setting it to `true` allows creation of truly one-jar deployment - plugin code and all dependencies are bundled into single file.

Undin (commented):
Actually, I think we don't need such complex configuration.
I can't imagine a situation when I want to use a custom jar task. What actually I want from gradle-intellij-plugin is to merge all jars related to plugin modules into a single jar since it's required by plugin model v2. And gradle-intellij-plugin can do it always (without any configuration) since there isn't any situation (at least, I don't know such cases) when it breaks something from IDE point of view

zielu:
Hi @Undin, let me share my use case. I have plugin consisting of several modules, one of purposes of modularisation is separating out implementations of APIs that are not compatible between IDE versions - branching in my opinion is a huge hassle to maintain. To load appropriate code version I use service loaders. I also run my plugin through an obfuscator and in order to hide modularisation the best way to do that is combine everything into single jar before obfuscation step.

So what I want from the plugin is combining all modules into a single jar, including service loader definitions and maybe including all external libraries too.

Undin:
I understood the idea but why do you need to provide a customization here at all? My suggestion is to **always** merge all jars from all modules in plugin project into single jar. It will solve your problem, common problem with plugin model v2, probably, even improve performance. And I don't see drawbacks of such solution

zielu:
Well, I'm lazy and wanted to have something working before 2.x arrives (#808 is marked as such). I don't want to solve issues with creating merged jar so decided to create a way to plug in existing solution with as little code as possible.

Thanks for review and comments, I understand that this hacky way is not beneficial to maintain in the plugin. 
At this time I'm not ready to invest any more time into this topic as I'm happy with working solution I wanted to share here.

PR #1092, opened 2022-08-22
Support unacceptable warnings

LChernigovskaya:


src/main/kotlin/org/jetbrains/intellij/tasks/VerifyPluginTask.kt
```diff
@@ -60,8 +60,13 @@ open class VerifyPluginTask @Inject constructor(
     fun verifyPlugin() {
         val creationResult = pluginDir.get().let { IdePluginManager.createManager().createPlugin(it.asFile.toPath()) }
         when (creationResult) {
-            is PluginCreationSuccess -> creationResult.warnings.forEach {
-                warn(context, it.message)
+            is PluginCreationSuccess -> {
+                creationResult.warnings.forEach {
+                    warn(context, it.message)
+                }
+                creationResult.unacceptableWarnings.forEach {
```
YannCebron, line 67: Technically these must not necessarily be "errors" in every case. E.g. a plugin that will never be uploaded to JB MP might not care about these problems. There must be way to suppress/configure this for such cases.

YannCebron (changes requested):

src/main/kotlin/org/jetbrains/intellij/tasks/VerifyPluginTask.kt
LChernigovskaya: Ok, will add. Should it be something like `ignoreUnacceptableWarnings` or `ignoreMarketplaceErrors`? The default value should be false, right?
Also, see that using the default value in name, description, etc. is considered an error. Should it be so? Or should we make it an unacceptable warning, so the vendors can suppress it also in the same way?

YannCebron: IMHO `ignoreUnacceptableWarnings`, default `false`
IMHO we can keep error level for all "default value" violations, even if they use plugin internally only, it makes sense to avoid these

YannCebron:
needs to be documented in SDK Docs after merge: https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#runpluginverifier-task

YannCebron:
where do we define what "unacceptable warnings" are as a concept? is this specific rating "hardcoded" in Plugin Verifier and could be changed depending on its version?

LChernigovskaya:
> where do we define what "unacceptable warnings" are as a concept? is this specific rating "hardcoded" in Plugin Verifier and could be changed depending on its version?

It's defined in intellij-plugin-structure in `IdePluginManager.createPlugin` method. It's necessary because some of the bundled (or old) plugins fail the checks which we decided to consider errors. We'd like to not allow the uploading of such plugins, but for dependencies, these plugins should be created successfully. It can be changed depending on intellij structure versions.

hsz (approved):
Thanks!

PR #1215, opened 2022-11-23
Make `Project.logCategory` appends `project.name` to `project.path` with a delimiter

bric3:
# Pull Request Details

In a nested multi module project, if the name and the path are not equal, they are appended however they are appended without any separator or delimiter. This change propose to insert a space before the name.

So
[gradle-intellij-plugin :modules:corecore] ...

Becomes
[gradle-intellij-plugin :modules:core core] ...

## Motivation and Context

This not a big fix, but this makes logs easier to read. 

## Types of changes

## Checklist

I don't think listing this change in the changelog noteworthy.

- [ ] My change requires a change to the [documentation](https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html).
- [ ] I have included my change in the [**CHANGELOG**](https://github.com/JetBrains/gradle-intellij-plugin/blob/master/CHANGES.md).
- [ ] All new and existing tests passed.

### DIFF
--- a/src/main/kotlin/org/jetbrains/intellij/utils.kt
+++ b/src/main/kotlin/org/jetbrains/intellij/utils.kt
@@ -283,7 +283,7 @@ private fun log(level: LogLevel, logCategory: String?, message: String, e: Throw
     }
 }
 
-fun Project.logCategory(): String = path + name.takeIf { ":$it" != path }.orEmpty()
+fun Project.logCategory(): String = path + if (path.subString(1) != name) " $name" else ""
 
 fun Task.logCategory(): String = project.logCategory() + path.removePrefix(project.logCategory())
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/jetbrains/intellij/utils.kt
```diff
@@ -283,7 +283,7 @@ private fun log(level: LogLevel, logCategory: String?, message: String, e: Throw
     }
 }
 
-fun Project.logCategory(): String = path + name.takeIf { ":$it" != path }.orEmpty()
+fun Project.logCategory(): String = path + if (path.subString(1) != name) " $name" else ""
```
bric3, line 286: Note I was wondering if using `endsWith` would be better, but I'm unsure about kotlin compatility.

```suggestion
fun Project.logCategory(): String = path + if (path.endsWith(name)) " $name" else ""
```

hsz:
Thanks!

bric3:
You're welcome thanks !

PR #1375, opened 2023-04-24
Hopefully fix build cacheability of `jarSearchableOptions`

martinbonnin:
_Preamble_: opening this PR as it's adding more context that an issue but I'm not 100% sure about everything that's going on in there. Feel free to close or modify this PR.

The issue is we have some tasks that are not cacheable because over overlapping outputs ([build scan](https://ge.apollographql.com/s/tstsfyflm5twe/timeline?cacheability=cacheable,overlapping-outputs,validation-failure&outcome=success,failed&sort=longest)). My understanding is that `outputDir` is the output of `BuildSearchableOptionsTask` and therefore the input of `JarSearchableOptionsTask`. The output of `JarSearchableOptionsTask` being decided by archiveBaseName ([here](https://github.com/JetBrains/gradle-intellij-plugin/blob/b7da6ab3b1cc309c0bf8fbf9f469277446c6ef1b/src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt#L934)).

Let me know what you think.


martinbonnin:
Looks like [CI is failing](https://github.com/JetBrains/gradle-intellij-plugin/actions/runs/4786677546/jobs/8511286535?pr=1375)? But not 100% sure why 

```
Error: HttpError: Resource not accessible by integration
``` 

Does that ring a bell?

hsz:
Thanks, Martin — this looks good to me!

PR #1438, opened 2023-07-23
Allow JBR runtime package of distributions to be used.

daym:
# Pull Request Details

Make JbrResolver use those gradle javaToolchains where the vendor is `JetBrains s.r.o`.

## Description

Make the JbrResolver also heed Jbr that are available in the gradle javaToolchains environment (even prefer those if they are set).

## Related Issue

https://github.com/JetBrains/gradle-intellij-plugin/issues/1437

## Motivation and Context

Sometimes, when making an IntelliJ IDEA plugin, it's not possible for the user to run the `runIde` or `buildSearchableOptions` tasks since those download a random JBR runtime and then try to run that, which fails.

It would be better to use the JBR that's available in the gradle javaToolchains already, if it is indeed available there.

## How Has This Been Tested

This was successfully tested by building a IntelliJ IDEA plugin from source on GNU Guix on Linux with this PR, after it had not worked on the same machine without the PR.

## Types of changes

## Checklist

- [x] My change requires a change to the [documentation](https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html).
- [ ] I have updated the documentation accordingly. ( https://plugins.jetbrains.com/docs/intellij/ide-development-instance.html#jetbrains-runtime-variants text in the `Gradle` tab needs to be updated. I don't know where the source is)
- [x] I have included my change in the [**CHANGELOG**](https://github.com/JetBrains/gradle-intellij-plugin/blob/master/CHANGELOG.md).


hsz:
Thank you, Danny! That's an excellent contribution!

daym:
Thanks!

I've also successfully tested your cleaned-up version the same way.

hsz:
Yesterday I sent PR to Gradle (gradle/gradle#26014), which makes JetBrains the _known JVM vendor_.

Since Gradle `8.4`, it'll be possible to use:

```
kotlin {
    jvmToolchain {
        languageVersion = JavaLanguageVersion.of(17)
        vendor = JvmVendorSpec.JETBRAINS
    }
}
```

hsz:
Also, it was necessary to change the vendor name from `JetBrains s.r.o` to just `JetBrains`.

If you already have JBR available locally, Gradle still matches it correctly.
However, suppose you expect to get it downloaded when using the [Foojay Toolchains Plugin](https://github.com/gradle/foojay-toolchains/tree/main). In that case, the `JetBrains` token will be correctly associated with JBR releases by the FooJay Disco API, but `JetBrains s.r.o` will not.

This change was also addressed with gradle/gradle#26022

daym:
> Since Gradle `8.4`, it'll be possible to use:
> 
> ```
> kotlin {
>     jvmToolchain {
>         languageVersion = JavaLanguageVersion.of(17)
>         vendor = JvmVendorSpec.JETBRAINS
>     }
> }
> ```

Nice! 

Would it also make sense to update the intellij platform plugin template's https://github.com/JetBrains/intellij-platform-plugin-template/blob/main/build.gradle.kts to contain this (eventually)? (If it's indeed intended to also *build* the plugin using JBR--not just run it using JBR)

PR #1519, opened 2023-12-30
Replace memoize internal Gradle API with cached provider using public API

TWiStErRob:
* Fixes https://github.com/JetBrains/gradle-intellij-plugin/issues/1517
 * Uses https://github.com/gradle/gradle/issues/25550

Testing:
 * Failing tests with Gradle 8.6-rc-1: https://github.com/TWiStErRob/gradle-intellij-plugin/actions/runs/7371154717?pr=1
 * Passing tests with `.cache()`: https://github.com/TWiStErRob/gradle-intellij-plugin/actions/runs/7365920095?pr=2


TWiStErRob:
@hsz this is ready for review. I copied @Vampire's workaround from [his Gradle issue](https://github.com/gradle/gradle/issues/25550) based on [conversation in Slack](https://gradle-community.slack.com/archives/CA745PZHN/p1703930171754399).
Then executed tests on 8.6-rc-1 in the CI matrix to ensure it solves the problem.
Please double-check that this solution has a `lazy { }` behavior as you would expect.

hsz:
This seems to fix the issue; thank you very much!

FTR, such a mechanism will be dropped in 2.0 as the dependencies management mechanism changes.

TWiStErRob:
Still need to use something until that :)

PR #1565, opened 2024-03-15
Split Mode

hsz:


hsz:
To runIde in split mode, you need:
```
tasks {
  runIde {
    splitMode = true
  }
}
```
But you can still keep the main runIde as it is and introduce new task based on it:
```
tasks {
  val runIdeInSplitMode by registering(RunIdeTask::class) {
    splitMode = true
  }
}
```
or even run PhpStorm in split mode:
```
tasks {
  val runIdeInSplitMode by registering(RunIdeTask::class) {
    type = IntelliJPlatformType.PhpStorm
    version = "241-EAP-SNAPSHOT"
    splitMode = true
  }
}
```
and local IDE works too!
```
tasks {
  val nightlyInSplitMode by registering(RunIdeTask::class) {
    localPath = file("/Users/hsz/Applications/IntelliJ IDEA Ultimate.app")
    splitMode = true
  }
}
```

src/main/kotlin/org/jetbrains/intellij/platform/gradle/argumentProviders/SplitModeArgumentProvider.kt
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Provides command line arguments for enabling Split Mode.
+ */
+class SplitModeArgumentProvider(
+    @Input val splitMode: Provider<Boolean>,
+) : CommandLineArgumentProvider {
+
+    override fun asArguments() = listOfNotNull(
+        "splitMode".takeIf { splitMode.get() },
```
chashnikov, line 17: This argument must be the first one. Are we sure that other providers won't execute before this one?

src/main/kotlin/org/jetbrains/intellij/platform/gradle/tasks/aware/SplitModeAware.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.jetbrains.intellij.platform.gradle.models.validateSupportedVersion
+import org.jetbrains.intellij.platform.gradle.utils.toVersion
+
+/**
+ * The interface provides the possibility to run IDE in Split Mode.
+ */
+interface SplitModeAware : IntelliJPlatformVersionAware {
+
+    /**
+     * Enables Split Mode when running the IDE.
```
chashnikov, line 18: We should provide an explanation what "Split Mode" means somewhere to make it visible to the user.
Also, it would be good to have it in the code as well, so I would copy the explanation from my pull request to KDoc of this class.

hsz: Documentation will be provided in SDK Docs and code — soon.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/argumentProviders/SplitModeArgumentProvider.kt
hsz: Thanks for highlighting that. I covered that scenario with https://github.com/JetBrains/intellij-platform-gradle-plugin/commit/38fd1c359d42b98c0c43d8537b139b1fd4a4154c#diff-053331baad74a29668f8bf6d9110d44e2f8a13d3118d39599eeaf13ec67654d1R63-R65

hsz:
Commits are cherry-picked due to the messy git log.

Issue #1599, opened 2024-04-24
2.0 - runIde task fails with "Cannot resolve 'product-info.json'" if `useBinaryReleases` feature is enabled


chashnikov:
### What happened?

I built IntelliJ Platform Gradle Plugin from sources which include `useBinaryReleases` feature (current 2.0.0-SNAPSHOT doesn't include it yet) and started `runIde` task. It failed with the following error message:
> Cannot resolve 'product-info.json' with: /home/nik/.gradle/caches/transforms-3/d89f9fc4990ceec8ab9fc252d20afe96/transformed/ideaIU-2024.1

The mentioned directory contains `idea-IU-241.14494.240` subdirectory with actual distribution inside. Switching off `useBinaryReleases` fixed the problem.

### Gradle IntelliJ Plugin version

2.0.0-SNAPSHOT

### Gradle version

8.5

### Operating System

Linux

hsz:
This has been fixed with https://github.com/JetBrains/intellij-platform-gradle-plugin/commit/bdfaf64f1729717b89bff14ee6b22a3fc492b9fa

Thanks for reporting.

PR #1600, opened 2024-04-24
BuildFeature: specify correct property name in KDoc

chashnikov:
[The documentation](https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-gradle-properties.html) mentions the proper prefix, but I was looking at the source code and spent some time trying to understand why I cannot disable binary releases (as a workaround for #1599).


hsz:
Thanks for spotting!

PR #1602, opened 2024-04-24
add lib/modules as directory containing plugin jars

mburyakov:
Platform module "com.intellij.modules.structuralsearch" is required for plugin "com.intellij.java". But "intellij.java.structuralSearch.jar" is located under "lib/modules". So "lib/modules" should be included into a classpath for test configuration.


hsz:
Thank you for highlighting that problem, Mihail!
I did some refactoring around collecting jars for plugins and bundled plugins, so that wouldn't be easy to merge your PR, but I copied relevant code. Thanks!

Issue #1563, opened 2024-03-15
Provide a way to run IDE with separate backend and frontend processes and have the plugin installed on the frontend side


chashnikov:
### Describe the need of your request

If you want to develop a plugin that should work in the frontend process (JetBrains Client), it'll be good to have a way to debug such a plugin without manually creating its distribution and installing it.

### Proposed solution

In #1562 a task which runs the backend part and the frontend part of the IDE in separate processes is introduced. By default, it includes the developed plugin in the backend process. Until we can automate determine whether the plugin is supposed to work on the backend side, on the frontend side or both, let's provide a configuration option to manually specify which process should have the plugin.

AlexPl292:
JFI, IdeaVim plugin works only on the client side, so this would be really helpful for us.

citizenmatt:
Is this planned for 2.0? It's the only reason IdeaVim would use split mode, so would be great if it can be in added sooner rather than later. Thanks!

chashnikov:
@citizenmatt I'm working on this, hope to have it done soon.

citizenmatt:
Cool. Thank you!

chashnikov:
This should be fixed in [2.0.0-beta2](https://github.com/JetBrains/intellij-platform-gradle-plugin/releases/tag/v2.0.0-beta2). Use the following configuration in build.gradle.kts:
```
intellijPlatform {
    splitMode = true
    splitModeTarget = SplitModeAware.SplitModeTarget.FRONTEND
}
```

chashnikov:
It appears that currently this doesn't work on macOS due to customized layout of IDE installation used by intellij-platform-gradle-plugin. I'll fix this on IDE side [RDCT-1375](https://youtrack.jetbrains.com/issue/RDCT-1375).

chashnikov:
The problem with installing plugin on macOS should be fixed if the next 2024.2 EAP build of the IDEs.

PR #1603, opened 2024-04-24
Add `targetProductPart` property to allow installing plugin on the frontend part when running in Split Mode (#1563)

chashnikov:
`targetProductPart` property is added to `IntelliJPlatformExtension` and `SplitModeAware`. It controls in which part of the IDE the developed plugin is installed. To do that, `PrepareSandboxTask` generates a properties file which configures the frontend process started from Split Mode. The path to the file is passed via `JETBRAINS_CLIENT_PROPERTIES` environment variable.

## Types of changes

## Checklist

- [x] My change requires a change to the [documentation](https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin.html).

### DIFF
--- a/src/main/kotlin/org/jetbrains/intellij/platform/gradle/argumentProviders/SandboxArgumentProvider.kt
+++ b/src/main/kotlin/org/jetbrains/intellij/platform/gradle/argumentProviders/SandboxArgumentProvider.kt
@@ -3,12 +3,15 @@
 package org.jetbrains.intellij.platform.gradle.argumentProviders
 
 import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputDirectory
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.OutputDirectory
 import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity.RELATIVE
 import org.gradle.process.CommandLineArgumentProvider
+import org.jetbrains.intellij.platform.gradle.tasks.RunIdeTask
 import org.jetbrains.intellij.platform.gradle.utils.asPath
 import java.io.File
 import java.nio.file.Path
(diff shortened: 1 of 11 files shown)


chashnikov:
I'm not quite sure about the property name (`targetProductPart`). If you have a better variant, let's discuss it.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/argumentProviders/SandboxArgumentProvider.kt
```diff
@@ -54,10 +57,36 @@ class SandboxArgumentProvider(
...
+    protected open fun computePluginPathProperties() = listOfNotNull(
         sandboxPluginsDirectory.ifExists { "-Didea.plugins.path=$it" },
         pluginPath?.let { "-Dplugin.path=$it" },
     )
 
-    private fun <T> DirectoryProperty.ifExists(block: (Path) -> T) = orNull?.asPath?.takeIf { it.exists() }?.run(block)
+    protected fun <T> DirectoryProperty.ifExists(block: (Path) -> T) = orNull?.asPath?.takeIf { it.exists() }?.run(block)
 }
+
+class SandboxArgumentProviderSplitModeAware(
```
chashnikov, line 71: I would prefer to avoid creating this subclass, but I don't know how to pass 'always false' property correctly when creating an instance where `splitMode` property is not available.

src/test/kotlin/org/jetbrains/intellij/platform/gradle/tasks/PrepareSandboxTaskTest.kt
```diff
@@ -361,6 +362,59 @@ class PrepareSandboxTaskTest : IntelliJPluginTestBase() {
         )
     }
 
+    @Test
+    fun `prepare sandbox for splitMode with plugin installed on frontend`() {
```
chashnikov, line 366: BTW if tests are written on JUnit5, I would simplify this code by using a parameterized test.

hsz: Good point — I'll move toward JUnit5 soon.

PR #1631, opened 2024-05-23
Update CONTRIBUTING.md

abrooksv:
Add note about the settings plugin usage since it requires placing `includeBuild` in a different location


hsz:
Thanks!

Issue #1779, opened 2024-09-30
Support dependency verification


AlexanderBartash:
### What happened?
Gradle [dependency verification](https://docs.gradle.org/current/userguide/dependency_verification.html) does not work with this plugin.

Multiple issues may be at play (see https://github.com/JetBrains/intellij-platform-gradle-plugin/issues/1779#issuecomment-2384461002 for correction & clarification):
1. If we signature verification i.e. `<verify-metadata>true</verify-metadata>` and `pgp` option in CLI. It fails with `Failed to create MD5 hash for file '/home/sasha/.gradle/caches/8.10.2/transforms/2f4561d46bc638866dfa4bcb9a564517/transformed/ideaIU-2023.3.8/plugins/JPA' as it does not exist.` with may or may not be a bug in Gradle.
2.  If we do not use signature verification and verify only hashes `<verify-metadata>false</verify-metadata>`  and no `pgp` option in CLI. Hashes for the actual artifacts are missing for dependencies created by this plugin, like `bundledPlugin:*`. Only hashes for ivy.xml are being generated by Gradle.
```
FAILURE: Build failed with an exception.

* What went wrong:
Configuration cache state could not be cached: field `__intellijPlatform__` of `org.jetbrains.intellij.platform.gradle.artifacts.transform.CollectorTransformer$Parameters` bean found in field `elements` of `org.gradle.internal.serialize.codecs.core.ResolutionBackedFileCollectionSpec` bean found in field `provider` of `org.gradle.internal.serialize.codecs.core.ProviderBackedFileCollectionSpec` bean found in field `element` of `java.util.Collections$SingletonList` bean found in field `elements` of `org.gradle.internal.serialize.codecs.core.ResolutionBackedFileCollectionSpec` bean found in field `__classpathSnapshot__` of `org.jetbrains.kotlin.gradle.tasks.KotlinCompile$ClasspathSnapshotProperties` bean found in field `__classpathSnapshotProperties__` of task `:compileKotlin` of type `org.jetbrains.kotlin.gradle.tasks.KotlinCompile`: error writing value of type 'org.gradle.api.internal.file.collections.DefaultConfigurableFileCollection'
> Dependency verification failed for configuration ':intellijPlatformDependency'
[...]

PR #1785, opened 2024-10-04
Fixed issue #1779 about compatibility with Gradle dependency verification. Previously it was failing with "Failed to create MD5 hash for file".

AlexanderBartash:
# Pull Request Details

I recommend to review this PR not by commits, but all of them together, because in the last few commits I have reverted a lot of code, which I added initially, while trying different approaches.

Combines #1782 #1783 & #1785 and adds another performance improvement.
Recommended to review by commits.

## Description

See the comment in `LocalIvyArtifactPathComponentMetadataRule.kt` file, it explains what was done pretty extensively.

~ExtractorTransformer now pre-creates Ivy XML files in the extracted IDE location, which makes it unnecessary to load all IDE plugins, resolve their location, create Ivy XML files on every build thereafter.~

~Potentially this is a breaking change because bundledPlugin & bundledModule dependencies now live in a separate repository created by `IntelliJPlatformDependenciesHelper#registerIntellijPlatformIvyRepo` That repository is not customizable by the user.~

## Related Issue

No issue.

## Motivation and Context

Life is short.

## How Has This Been Tested

Tests, manually. On my project also on https://github.com/InSyncWithFoo/ryecharm

## Types of changes

- [x ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x ] I have read the [**CONTRIBUTING**](https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/master/CONTRIBUTING.md) document.
- [ x] My code follows the code style of this project.
- [x ] I have included my change in the [**CHANGELOG**](https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/master/CHANGELOG.md).
- [ x] All new and existing tests passed.

### DIFF
--- a/src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
+++ b/src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
@@ -13,6 +13,7 @@ import org.gradle.api.artifacts.ProjectDependency
 import org.gradle.api.artifacts.dsl.DependencyHandler
 import org.gradle.api.file.Directory
 import org.gradle.api.file.ProjectLayout
+import org.gradle.api.initialization.resolve.RulesMode
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Provider
 import org.gradle.api.provider.ProviderFactory
@@ -61,6 +62,7 @@ class IntelliJPlatformDependenciesHelper(
     private val objects: ObjectFactory,
     private val providers: ProviderFactory,
     private val rootProjectDirectory: Path,
+    private val metadataRulesModeProvider: Provider<RulesMode>
 ) {
 
     private val log = Logger(javaClass)
@@ -708,7 +710,7 @@ class IntelliJPlatformDependenciesHelper(
                     module = type.code,
                     revision = version,
                 ),
-                publications = listOf(artifactPath.toIvyArtifact()),
+                publications = listOf(artifactPath.toAbsolutePathIvyArtifact()),
             )
         }
 
@@ -771,7 +773,7 @@ class IntelliJPlatformDependenciesHelper(
                     module = id,
                     revision = version,
                 ),
-                publications = listOf(artifactPath.toIvyArtifact()),
+                publications = artifactPath.toIvyArtifacts(metadataRulesModeProvider, platformPath.get()),
                 dependencies = plugin.collectDependencies(),
             )
         }
@@ -822,7 +824,7 @@ class IntelliJPlatformDependenciesHelper(
                     writeIvyModule(group, name, version, artifactPath) {
                         IvyModule(
                             info = IvyModule.Info(group, name, version),
-                            publications = listOf(artifactPath.toIvyArtifact()),
+                            publications = artifactPath.toIvyArtifacts(metadataRulesModeProvider, platformPath.get()),
                             dependencies = plugin.collectDependencies(alreadyProcessedOrProcessing + id),
                         )
                     }
@@ -858,7 +860,9 @@ class IntelliJPlatformDependenciesHelper(
         val group = Dependencies.BUNDLED_MODULE_GROUP
         val version = productInfo.get().buildNumber
         val platformPath = platformPath.get()
-        val artifacts = classPath.map { platformPath.resolve(it).toIvyArtifact() }
+        val artifacts = classPath.flatMap {
+                path -> platformPath.resolve(path).toIvyArtifacts(metadataRulesModeProvider, platformPath)
+        }
 
         /**
          * For bundled modules, usually we don't have a path to their archive (jar), since we get them from [ProductInfo.layout], which does not have a path.
@@ -905,7 +909,7 @@ class IntelliJPlatformDependenciesHelper(
                     module = name,
                     revision = version,
                 ),
-                publications = listOf(artifactPath.toIvyArtifact()),
+                publications = listOf(artifactPath.toAbsolutePathIvyArtifact()),
             )
         }
 
@@ -946,7 +950,7 @@ class IntelliJPlatformDependenciesHelper(
                     module = name,
                     revision = version,
                 ),
-                publications = listOf(artifactPath.toIvyArtifact()),
+                publications = listOf(artifactPath.toAbsolutePathIvyArtifact()),
             )
         }
 
--- a/src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
+++ b/src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
@@ -62,7 +62,7 @@ class IntelliJPlatformDependenciesHelper(
     private val objects: ObjectFactory,
     private val providers: ProviderFactory,
     private val rootProjectDirectory: Path,
-    private val metadataRulesModeProvider: Provider<RulesMode>
+    private val metadataRulesModeProvider: Provider<RulesMode>,
 ) {
 
     private val log = Logger(javaClass)
@@ -861,7 +861,7 @@ class IntelliJPlatformDependenciesHelper(
         val version = productInfo.get().buildNumber
         val platformPath = platformPath.get()
         val artifacts = classPath.flatMap {
-                path -> platformPath.resolve(path).toIvyArtifacts(metadataRulesModeProvider, platformPath)
+            platformPath.resolve(it).toIvyArtifacts(metadataRulesModeProvider, platformPath)
         }
 
         /**
--- a/src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformRepositoriesHelper.kt
+++ b/src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformRepositoriesHelper.kt
@@ -3,7 +3,6 @@
 package org.jetbrains.intellij.platform.gradle.extensions
 
 import org.gradle.api.artifacts.dsl.RepositoryHandler
-import org.gradle.api.artifacts.repositories.IvyArtifactRepository
 import org.gradle.api.credentials.HttpHeaderCredentials
 import org.gradle.api.credentials.PasswordCredentials
 import org.gradle.api.flow.FlowProviders
@@ -14,7 +13,7 @@ import org.gradle.api.provider.ProviderFactory
 import org.gradle.authentication.http.HttpHeaderAuthentication
 import org.gradle.internal.os.OperatingSystem
 import org.gradle.kotlin.dsl.*
-import org.jetbrains.intellij.platform.gradle.Constants.Configurations.Dependencies
+import org.jetbrains.intellij.platform.gradle.Constants.Configurations
 import org.jetbrains.intellij.platform.gradle.CustomPluginRepositoryType
 import org.jetbrains.intellij.platform.gradle.GradleProperties
 import org.jetbrains.intellij.platform.gradle.artifacts.repositories.PluginArtifactRepository
@@ -23,8 +22,7 @@ import org.jetbrains.intellij.platform.gradle.get
 import org.jetbrains.intellij.platform.gradle.services.ShimManagerService
 import java.net.URI
 import java.nio.file.Path
-import kotlin.io.path.absolute
-import kotlin.io.path.invariantSeparatorsPathString
+import kotlin.io.path.pathString
 
 private const val SHIM_MANAGER = "shimManager"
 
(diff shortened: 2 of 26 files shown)


src/main/kotlin/org/jetbrains/intellij/platform/gradle/artifacts/transform/ExtractorTransformer.kt
```diff
@@ -105,17 +115,76 @@ abstract class ExtractorTransformer @Inject constructor(
                     execOperations.exec {
                         commandLine("hdiutil", "detach", "-force", "-quiet", tempDirectory)
                     }
+
+                    createIvyXmls(tempDirectory)
```
AlexanderBartash, line 119: I am not 100% sure if this is correct.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
```diff
@@ -343,13 +340,69 @@ class IntelliJPlatformDependenciesHelper(
...
+            "Ivy IntelliJ Platform Artifacts Repository",
+            repositories,
+            providers.localPlatformArtifactsPath(ivyLocation),
+            platformPath.get()
+        ).content {
+            includeGroup(Dependencies.BUNDLED_MODULE_GROUP)
+            includeGroup(Dependencies.BUNDLED_PLUGIN_GROUP)
+        }
+    }
+
```
AlexanderBartash, line 405: I could create it in `createDefaultRepositores()` as well, so that users have a dedicated method for this, but I thought it would be bad, because then, there we would be updating ivy & artifact path of that repo on the fly. Which is error prone, if this is exposed to customizations.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
```diff
@@ -812,65 +899,147 @@ class IntelliJPlatformDependenciesHelper(
...
-                    )
+    companion object {
+        private val log = Logger(IntelliJPlatformDependenciesHelper::class.java)
+
+        internal fun Path.getBundledPlugins(pluginManager: IdePluginManager): Map<String, IdePlugin> {
+            return this.resolve("plugins")
+                .listDirectoryEntries()
+                .filter { it.isDirectory() }
+                .mapNotNull { path ->
+                    // TODO: try not to parse all plugins at once
```
AlexanderBartash, line 910: Now it parses them only if `local("/opt/ideaIU")` is used. Otherwise it is done in the transformer only once.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/artifacts/transform/ExtractorTransformer.kt
```diff
@@ -105,17 +115,76 @@ abstract class ExtractorTransformer @Inject constructor(
...
+        log.info("Creating an Ivy repository in '$platformPath'.")
+
+        val writtenIvyModules = HashSet<String>()
+        val productInfo = platformPath.productInfo()
+        val version = productInfo.buildNumber
+
+        val pluginManager = IdePluginManager.createManager(createTempDirectory())
+        val plugins = platformPath.getBundledPlugins(pluginManager)
+
+        for (plugin in plugins.values) {
```
AlexanderBartash, line 155: I think some optimizations are possible here for caching & skipping what was already discovered, but I do not want to over complicate it. Also since it is done in the transformer now, it does not matter that much.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
```diff
@@ -812,65 +899,147 @@ class IntelliJPlatformDependenciesHelper(
...
+            productInfo: ProductInfo,
+            platformPath: Path,
+            bundledPlugins: Map<String, IdePlugin>,
+
+            writtenIvyModules: MutableSet<String>,
+            providers: ProviderFactory,
+            pathToWriteInto: Path,
+
+            isCreateModules: Boolean
+        ): List<IvyModule.Dependency> {
```
AlexanderBartash, line 936: So the above list of args is not something I am proud of. It is like that because I've kept things as they were as much as possible to avoid making this PR too big. It can be refactored later if needed.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
```diff
@@ -812,65 +899,147 @@ class IntelliJPlatformDependenciesHelper(
...
+        internal fun writeIvyModule(
+            group: String,
+            artifact: String,
+            version: String,
+
+            writtenIvyModules: MutableSet<String>,
+            providers: ProviderFactory,
+            pathToWriteInto: Path,
+
+            block: () -> IvyModule
```
AlexanderBartash, line 1022: The same story.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformRepositoriesHelper.kt
```diff
@@ -142,33 +143,91 @@ class IntelliJPlatformRepositoriesHelper(
...
+            artifactLocationPath: Path?,
+            action: IvyRepositoryAction = {},
+        ): IvyArtifactRepository {
+            val repository = createIvyArtifactRepository(
+                repositoryName, repositories, ivyLocationPath, artifactLocationPath
+            )
+
+            // For performance reasons make it exclusive.
+            // https://docs.gradle.org/current/userguide/declaring_repositories_adv.html#declaring_content_exclusively_found_in_one_repository
+            repositories.exclusiveContent {
```
AlexanderBartash, line 179: So this is also an improvement over #1783 I noticed that there is a dedicated API for telling Gradle "look for this group only here" which is probably the safest option possible, especially if combined with prefixed groups with "com.jetbrains".

AlexanderBartash: Btw, if we pre-create that new dynamic Ivy repository here as well, we can also make it exclusive, it may save a couple of seconds on the build. But again, I am not sure that creating a repo with broken paths is something we should be doing.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/models/IvyModule.kt
```diff
@@ -88,10 +100,86 @@ data class IvyModule(
...
+    // is marked as optional.
+    // In this case, it should be ok to have an absolute path in the name, because this is a fallback method,
+    // which called for directories only (at least at the moment of writing this).
+    // See comments in explodeIntoIvyJarsArtifacts on why having abs path in name may be bad.
+    return IvyModule.Artifact(type = alwaysEmpty, name = trimmedAbsolutePath, ext = ext)
+}
+
+internal fun Path.toBundledIvyArtifacts(basePath: Path): List<IvyModule.Artifact> {
+    return this.explodeIntoIvyJarsArtifacts(basePath)
+}
```
AlexanderBartash, line 122: If you noticed, previously I introduced here two separate methods:
```
internal fun Path.toBundledModuleIvyArtifacts(): List<IvyModule.Artifact> {
    return this.explodeIntoIvyJarsArtifacts()
}

internal fun Path.toLocalPluginIvyArtifacts(): List<IvyModule.Artifact> {
    return this.explodeIntoIvyJarsArtifacts()
}
```

Eventually I have merged them into one.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/artifacts/transform/ExtractorTransformer.kt
```diff
@@ -116,6 +130,60 @@ abstract class ExtractorTransformer @Inject constructor(
...
     }
 
+    /**
+     * Pre-create Ivy XML files so that later this directory can be used as an Ivy repository without having to discover
+     * modules & plugins each time the build is run.
+     */
+    private fun createIvyXmls(platformPath: Path) {
+        val isIde = platformPath.listDirectoryEntries().map { it.name }.containsAll(
+            listOf("bin", "lib", "plugins", "product-info.json")
+        )
```
AlexanderBartash, line 140: Is this correct? It works for IU & IC.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformRepositoriesHelper.kt
```diff
@@ -142,33 +146,156 @@ class IntelliJPlatformRepositoriesHelper(
...
+            artifactLocationPath: Path?
+        ) = repositories.ivy {
+            name = repositoryName
+
+            // The contract is that we are working with absolute normalized paths here.
+            val absNormIvyLocationPath = ivyLocationPath.absolute().normalize()
+            val absNormArtifactLocationPath = artifactLocationPath?.absolute()?.normalize()
+
+            // Location of Ivy files generated for the current project.
+            val ivyPath = providers.localPlatformArtifactsPath(absNormIvyLocationPath)
```
AlexanderBartash, line 230: `localPlatformArtifactsPath` does not make much sense with the dynamic repository. Should I apply it only for repositories which are static?

src/main/kotlin/org/jetbrains/intellij/platform/gradle/models/IvyModule.kt
```diff
@@ -87,11 +100,117 @@ data class IvyModule(
...
+ *
+ * Also, if the optional basePath is given, the path will be relative to it. In this case it also does not have the
+ * leading path separator.
+ *
+ * @see org.jetbrains.intellij.platform.gradle.extensions.IntelliJPlatformRepositoriesHelper.createIvyArtifactRepository
+ */
+private fun Path.containingDirPathStringRelativeTo(basePath: Path? = null): String {
+    // The contract is that we are working with absolute normalized paths here.
+    val absNormalizedPath = this.absolute().normalize()
+    val absNormalizedBasePath = basePath?.absolute()?.normalize()
```
AlexanderBartash, line 184: So, I do this very often. The reason is that I try to design functions here as independent black boxes, which should just work by themselves. It should be more fool-proof this way.

AlexanderBartash:
I think I have finally made it all work, but it seems that tests on windows fail randomly. I've seen this error first on Gradle 8.2 Win, then on Gradle 8.10.2 Win, now it is back on 8.2 Win.

```
org.gradle.testkit.runner.UnexpectedBuildFailure: Unexpected build execution failure in C:\Users\RUNNER~1\AppData\Local\Temp\tmp12965048409070354875 with arguments [-Dorg.gradle.kotlin.dsl.scriptCompilationAvoidance=false, verifyPluginProjectConfiguration, --stacktrace, --configuration-cache]

Output:
Calculating task graph as no configuration cache is available for tasks: verifyPluginProjectConfiguration

> Configure project :
Kotlin DSL property assignment is an incubating feature.
e: file:///C:/Users/runneradmin/AppData/Local/Temp/tmp12965048409070354875/build.gradle.kts:1:1: Cannot access implicit script receiver class 'org.gradle.api.Project'. Check your module classpath for missing or conflicting dependencies
e: file:///C:/Users/runneradmin/AppData/Local/Temp/tmp12965048409070354875/build.gradle.kts:8:1: Function invocation 'version(...)' expected
e: file:///C:/Users/runneradmin/AppData/Local/Temp/tmp12965048409070354875/build.gradle.kts:8:1: Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
public infix fun PluginDependencySpec.version(version: String?): PluginDependencySpec defined in org.gradle.kotlin.dsl
public infix fun PluginDependencySpec.version(version: Provider<String>): PluginDependencySpec defined in org.gradle.kotlin.dsl
e: file:///C:/Users/runneradmin/AppData/Local/Temp/tmp12965048409070354875/build.gradle.kts:16:1: Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
public fun Project.kotlin(configure: Action<KotlinJvmProjectExtension>): Unit defined in org.gradle.kotlin.dsl
public fun DependencyHandler.kotlin(module: String, version: String? = ...): Any defined in org.gradle.kotlin.dsl
public fun SourceSet.kotlin(configure: Action<SourceDirectorySet>): Unit defined in org.gradle.kotlin.dsl
public fun PluginDependenciesSpec.kotlin(module: String): PluginDependencySpec defined in org.gradle.kotlin.dsl
```

src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformRepositoriesHelper.kt
AlexanderBartash: I did that. Now the dynamic repo stores Ivy.xml files into "localPlatformArtifacts" inside `platformPath` (not customisable). If a `local("/path/to/ide")` is used, the customizable location is used from `providers.localPlatformArtifactsPath(absNormIvyLocationPath)`

AlexanderBartash:
I am done with this PR, I do not plat to change anything, unless there is a feedback. I tested this very extensively on my project on both linux and windows. Also I used this to build https://github.com/InSyncWithFoo/ryecharm on linux, it works.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/models/IvyModule.kt
```diff
@@ -71,7 +75,16 @@ data class IvyModule(
  * <ivy-module version="2.0">
  *     ...
  *     <publications>
- *         <artifact conf="default" ext="jar" name="/path/to/artifact.jar" type="jar" />
```
AlexanderBartash, line 74: Previously in all paths we had the leading slash, but the artifact pattern was already `/[artifact]` which would lead to double slash eventually. It seems to work like that, but I fixed this anyway.

AlexanderBartash:
Rebased on HEAD. Also reverted one change:
> Revert prefixing local groups with "com.jetbrains.localhost-only" because group names are hardcoded in the IDE https://github.com/JetBrains/intellij-community/blob/c8ba8401f73488f966b5ed5b5c5afaa38a9bcbdf/plugins/devkit/intellij.devkit.gradle/src/IntelliJPlatformAttachSourcesProvider.kt#L64-L67 AlexanderBartash 5 minutes ago f0bb8cc0

AlexanderBartash:
Rebased on HEAD. Also reverted one change:
> Revert prefixing local groups with "com.jetbrains.localhost-only" because group names are hardcoded in the IDE https://github.com/JetBrains/intellij-community/blob/c8ba8401f73488f966b5ed5b5c5afaa38a9bcbdf/plugins/devkit/intellij.devkit.gradle/src/IntelliJPlatformAttachSourcesProvider.kt#L64-L67 AlexanderBartash 5 minutes ago f0bb8cc0

src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
AlexanderBartash: I actually think now, that this approach is better than what was before, because previously (I have not actually tried it) the developer did not call `createDefaultRepositores()` these dependencies won't work, Which is counter intuitive, why we allow defining dependencies which will not work. With this approach if a dependency on a bundled plugin or module is added it will work.

AlexanderBartash:
Rebased.

AlexanderBartash:
Rebased on the latest, including the not yet merged #1785 (because tests in master are broken).
Also created a new implementation for resolving relative paths, because apparenyly we have to support `RepositoriesMode.FAIL_ON_PROJECT_REPOS`. The essence of it is described in https://github.com/JetBrains/intellij-platform-gradle-plugin/pull/1785/commits/02396acc8aab915c216f73013c54eeafa05e9bec#diff-550b94ce513e8351a885e962947d583d45c59187d91264e2f062f583b5cd5bce

AlexanderBartash:
Fixed issues on Windows. The failed tests on OSX are not related to this PR, see https://github.com/JetBrains/intellij-platform-gradle-plugin/pull/1792#issuecomment-2408716496

src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
```diff
@@ -999,20 +1002,75 @@ class IntelliJPlatformDependenciesHelper(
...
+        // It is actually could be also "17.0.6-b829.4"
+        // https://github.com/JetBrains/JetBrainsRuntime/releases?page=8
+        // The latest are without -
+        // https://github.com/JetBrains/JetBrainsRuntime/releases
+        val fixedVersion = "${javaVersion}b${buildVersion}"
+
+        val isJcef = implementorVersion.endsWith("-jcef") || implementorVersion.endsWith("-JCEF")
+        val variant = if (isJcef) "jcef" else null
+
+        return@cachedProvider buildJetBrainsRuntimeVersion(fixedVersion, variant)
```
AlexanderBartash, line 1073: This change is from #1792 because I needed to have working tests.

AlexanderBartash:
Pushed 1 commit because I realized that there is a setting `RulesMode.FAIL_ON_PROJECT_RULES` just like `RepositoriesMode.FAIL_ON_PROJECT_REPOS`. The commit message pretty much explains it:

> Refactored the logic to use the relative paths only if RulesMode.PREFER_PROJECT mode is used, because we can not use a project level ComponentMetadataRule with RulesMode.PREFER_SETTINGS or RulesMode.FAIL_ON_PROJECT_RULES. Added some logging and comments. Added new integration tests to make sure RepositoriesMode.FAIL_ON_PROJECT_REPOS & RulesMode.FAIL_ON_PROJECT_RULES are supported. AlexanderBartash 2 minutes ago a47b0b4a

PR #1792, opened 2024-10-12
An attempt to fix bugs & broken tests.

AlexanderBartash:
# Pull Request Details

An attempt to fix broken tests in master.

## Description

:point_up: 

## Related Issue

No.

## Motivation and Context

They are broken?

## How Has This Been Tested

By running tests.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x ] I have read the [**CONTRIBUTING**](https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/master/CONTRIBUTING.md) document.
- x[ ] My code follows the code style of this project.
- [ ] My change requires a change to the [documentation](https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin.html).
- [ ] I have updated the documentation accordingly.
- [ ] I have included my change in the [**CHANGELOG**](https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/master/CHANGELOG.md).
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1804, opened 2024-10-17
Fixed generation of duplicate files in ".intellijPlatform/localPlatformArtifacts" with different version numbers.

AlexanderBartash:
# Pull Request Details

Fixed generation of duplicate files in ".intellijPlatform/localPlatformArtifacts" with different version numbers.
As usual, it also includes #1792 because master is broken. The tests will fail on OSX, see why in #1792

## Description

The duplicates were created because here it was `val version = requireNotNull(plugin.pluginVersion)`
https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/main/src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt#L814

So I did the same in `createIntelliJPlatformBundledPlugin`.

This also results in a more consistent dependency graph, e.g. see before & after:
image

## Related Issue

No.

## Motivation and Context

It was confusing.

## How Has This Been Tested

Running tests, manually.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x ] I have read the [**CONTRIBUTING**](https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/master/CONTRIBUTING.md) document.
- [ x] My code follows the code style of this project.
- [ x] I have included my change in the [**CHANGELOG**](https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/master/CHANGELOG.md).
- [x ] All new and existing tests passed.

### DIFF
--- a/src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
+++ b/src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
@@ -741,7 +741,7 @@ class IntelliJPlatformDependenciesHelper(
         }
 
         val artifactPath = requireNotNull(plugin.originalFile)
-        val version = getVersionForBundledPlugin(plugin)
+        val version = requireNotNull(plugin.pluginVersion)
 
         writeIvyModule(Dependencies.BUNDLED_PLUGIN_GROUP, id, version) {
             IvyModule(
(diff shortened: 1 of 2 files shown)


AlexanderBartash:
Tests on OSX failed because of #1792 and on Windows that is a random test failure, which seems to happen in this project.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
```diff
@@ -741,7 +741,7 @@ class IntelliJPlatformDependenciesHelper(
         }
 
         val artifactPath = requireNotNull(plugin.originalFile)
-        val version = baseVersion.orElse(productInfo.map { it.version }).get()
+        val version = getVersionForBundledPlugin(plugin)
```
hsz, line 744: Nice.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
```diff
@@ -793,6 +793,12 @@ class IntelliJPlatformDependenciesHelper(
         )
     }
 
+    // These two methods were created to make clearly visible that we want to use consistent version numbers in
+    // collectDependencies as well as in createIntelliJPlatformBundledModule & createIntelliJPlatformBundledPlugin
+    // When the version is retrieved like this, it is impossible to change it one place & forger to chang another.
+    private fun getVersionForBundledModule() = productInfo.get().buildNumber
+    private fun getVersionForBundledPlugin(plugin: IdePlugin) = requireNotNull(plugin.pluginVersion)
```
hsz, line 800: This method seems redundant, as its usage always appears next to other statements like `requireNotNull(plugin.PROPERTY)`, so it's clear value should come from `plugin` object.

src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt
```diff
@@ -793,6 +793,12 @@ class IntelliJPlatformDependenciesHelper(
         )
     }
 
+    // These two methods were created to make clearly visible that we want to use consistent version numbers in
+    // collectDependencies as well as in createIntelliJPlatformBundledModule & createIntelliJPlatformBundledPlugin
+    // When the version is retrieved like this, it is impossible to change it one place & forger to chang another.
+    private fun getVersionForBundledModule() = productInfo.get().buildNumber
```
hsz, line 799: Given that it's indeed easy to make a mistake here, this was a good idea. However, given we have that method already, I hid the whole Ivy XML creation in such a method.

PR #1842, opened 2024-12-13
Suppress messages that indicate missing Platform layout entries in the filesystem

novotnyr:
# Pull Request Details

Logs might contain the following messages:

```
Layout component 'emoji-picker' has some nonexistent 'classPath' elements: 'plugins/emojipicker.jar'
```

This is an incorrect Platform layout where the declaration in the `product-info.json` does not match the real filesystem of the Platform. For example, `emoji-picker` is declared in the macOS, but available only on Linux.

Such messages are not actionable from the plugin developer side and should be suppressed.

When constructing Platform model, ignore missing layout entries, too.

Note that such message might occur when running the Plugin Verifier via `verifyPlugin` task. Use the `freeArgs()` and the `-missing-layout-classpath-file` CLI switch for the Plugin Verifier.

## Description

- Use a customized way to create an IDE for the plugin model in the IntelliJ helper.

## Motivation and Context

Misleading and nonactionable messages confuse plugin developers and make them think this should be fixed on their side

## How Has This Been Tested

Ran against sample project based on the IntelliJ Plugin Template.

## Types of changes

## Checklist

- [ ] I have included my change in the [**CHANGELOG**](https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/master/CHANGELOG.md).

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 # libraries
 annotations = "26.0.1"
-intellijStructure = "3.293"
+intellijStructure = "3.294"
 intellijPluginRepositoryRestClient = "2.0.39"
 kotlinxSerializationJson = "1.6.3"
 okio = "1.17.6"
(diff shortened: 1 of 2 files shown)


hsz:
Thanks!

PR #2067, opened 2025-12-09
Fix the execution of JavaExec based tasks when the class path gets shortened

garidozv:
# Pull Request Details

For `JavaExec` tasks Gradle has to shorten the command line if it gets over the limit (either defined by OS or by a system property). This is currently done by shortening the class path and creating a pathing jar that will contain the whole class path in its `Class-Path` metadata entry and that will be the only jar directly passed as part of the command. 

The [PathClassLoader](https://github.com/JetBrains/intellij-community/blob/406e37c1085bd873fea7ac00aa66822639cc8457/platform/boot/src/com/intellij/util/lang/PathClassLoader.java#L28) supports this, but only if a specific system property has been set (see [this](https://github.com/JetBrains/intellij-community/blob/406e37c1085bd873fea7ac00aa66822639cc8457/platform/boot/src/com/intellij/util/lang/PathClassLoader.java#L118)).

## Description

This PR sets the necessary property to enable successful task execution in cases where the class path gets shortened by using the pathing JAR and adds an integration test to cover this case.

The property will be set  unconditionally just for the task that are both `RunnableIdeAware` and `JavaExec`.
Maybe some sort of check should be introduced here, but even if the class path does end up having only one real JAR (which shouldn't happen for these tasks), there are necessary checks in `PathClassLoader` that will make sure it is handled properly.

## Motivation and Context

Currently, Gradle uses pathing jar to shorten the class path in cases where the command line gets too long. 
There is an open [PR](https://github.com/gradle/gradle/pull/34227) that will introduce the use of args file in these cases and avoid class path shortening for Java 9+. This change will also solve this problem.

## How Has This Been Tested

I've created a single integration test to cover this case in a separate class and with a separate testing resources.
Since the fix relies on a `PathClassLoader` functionality that has been introduced in version `2025.3`, the test targets that version as well. The existing caching setup wasn't working properly with this new version, so I had to disable it.

Although the existence of this new system property won't do any harm in cases where the class path shortening doesn't happen, maybe a test should be added for that case as well?

I wasn't able to run tests locally until I edited [these](https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/0a708dc13018208734c285b06e0f08bf388e947e/src/main/kotlin/org/jetbrains/intellij/platform/gradle/tasks/tasks.kt#L330) properties to use `.orNull ?: ""` instead. This change is not included in the PR as I assumed it has something to do with the environment I'm running in and might not be an actual problem.

## Types of changes

## Checklist

- [ ] All new and existing tests passed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,7 +76,7 @@ jobs:
 
       # Store an already-built plugin as an artifact for downloading
       - name: Upload artifact
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           path: ./build/libs/
 
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -87,6 +87,6 @@ jobs:
 
       # Store already-built plugin as an artifact for downloading
       - name: Upload artifact
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           path: ./build/libs/*-SNAPSHOT.jar
--- a/.github/workflows/reusable-codeInspection.yml
+++ b/.github/workflows/reusable-codeInspection.yml
@@ -14,5 +14,5 @@ jobs:
           fetch-depth: 0 # only report issues that appeared in a PR
 
       - name: Qodana - Code Inspection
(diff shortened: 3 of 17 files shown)


hsz:
Thanks!

PR #2085, opened 2026-01-28
Tiered name-conflict handling in PrepareSandboxTask (pre-detect, hash when needed, rename only for real conflicts)

arjav1528:
# Pull Request Details

**Tiered name-conflict handling in PrepareSandboxTask (pre-detect, hash when needed, rename only for real conflicts)**

## Description

Adds a tiered approach to jar name conflicts in `PrepareSandboxTask`:

1. **Pre-detect** — `detectNameConflicts()` builds a map of filename → list of `File` from `runtimeClasspath`, `pluginJar`, and `pluginsClasspath` and returns only entries with more than one file. No extra file I/O.
2. **Resolve** — `resolveNameConflicts()` runs only when there are exactly two files with the same name: if their MD5 hashes match, it logs "Skipping duplicate content" and does not treat them as a conflict; otherwise it logs "Name conflict detected" with paths.
3. **Rename** — `nameCollisionHelper` uses `nameConflicts` and runs the existing `core` / `core_1` / `core_2` renaming loop only when `conflictFiles != null && conflictFiles.size > 1`.

New helpers: `detectNameConflicts()`, `resolveNameConflicts()`, and `quickFileHash()`. A `nameConflicts` field is set at the start of `copy()` and read by `nameCollisionHelper`. Conflict detection and resolution run at **execution time** (in `copy()` before `super.copy()`), not during task configuration, to avoid resolving `runtimeClasspath` / `pluginsClasspath` too early and triggering "Cannot mutate the state of configuration ':api' after the configuration's child configuration ':intellijPlatformRuntimeClasspath' was resolved."

## Motivation and Context

- **Performance** — Hashing is done only for two-file conflicts; renaming is applied only when there is a real conflict.
- **Correctness** — Same-name jars with identical content are no longer treated as conflicts that need renaming.
- **Configuration safety** — Moving detection/resolution to execution time avoids mutating Gradle configurations during the configuration phase and fixes the `:intellijPlatformRuntimeClasspath` / `:api` resolution error in projects using the configuration cache.

## How Has This Been Tested

- **`rename jars with same names`** — Confirms three `core.jar` files still produce `core.jar`, `core_1.jar`, and `core_2.jar`.
- **`PrepareSandboxTaskTest`** — Full class run: all tests pass.
- **Environment** — Gradle 9.3.0, Kotlin (embedded), macOS. Run with:
  - `./gradlew test --tests "org.jetbrains.intellij.platform.gradle.tasks.PrepareSandboxTaskTest.rename jars with same names"`
  - `./gradlew test --tests "org.jetbrains.intellij.platform.gradle.tasks.PrepareSandboxTaskTest"`

## Types of changes

## Checklist

### DIFF
--- a/src/main/kotlin/org/jetbrains/intellij/platform/gradle/tasks/PrepareSandboxTask.kt
+++ b/src/main/kotlin/org/jetbrains/intellij/platform/gradle/tasks/PrepareSandboxTask.kt
@@ -33,6 +33,8 @@ import org.jetbrains.intellij.platform.gradle.utils.asPath
 import org.jetbrains.intellij.platform.gradle.utils.extensionProvider
 import org.jetbrains.intellij.platform.gradle.utils.safePathString
 import org.jetbrains.kotlin.gradle.utils.named
+import java.nio.file.Path
+import java.security.MessageDigest
 import kotlin.io.path.*
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/jetbrains/intellij/platform/gradle/tasks/PrepareSandboxTask.kt
```diff
@@ -33,6 +33,7 @@ import org.jetbrains.intellij.platform.gradle.utils.asPath
 import org.jetbrains.intellij.platform.gradle.utils.extensionProvider
 import org.jetbrains.intellij.platform.gradle.utils.safePathString
 import org.jetbrains.kotlin.gradle.utils.named
+import java.io.File
```
hsz: Please do not work on `File` but `Path` instead.

PR #2103, opened 2026-03-16
Add dedicated split-mode backend and frontend run tasks

DenisZaichenko:
- introduce execution-mode-aware runIde wiring for backend and frontend launches
- launch the frontend as JetBrains Client using frontend.properties and persisted join links

With these changes, we can:

debug the backend and the frontend independently
create compose IDE run configurations for coordinated backend/frontend launches
configure frontend settings in the same way as backend settings
see frontend startup status and logs directly in the IDE as a dedicated run target


jreznot:
Does front-end launch depend on the order of start? We need it to await until backend appears with a timeout.

PR #2164, opened 2026-06-18
Make split-mode IDE runs observable and one-command ready

kirillk:
## TL;DR

This PR makes IntelliJ split-mode runs observable, agent-ready, and easier to start from one place.

Before this work, split-mode failures could leave users and automation without clear evidence: no reliable PID, unclear effective sandbox paths, ambiguous port ownership, stale backend state, and separate backend/frontend startup steps that were easy to get wrong. Debug launches also exposed a gap where the combined frontend process could inherit the backend JDWP configuration and fail before the JetBrains Client main class started.

This PR closes those gaps by making the runtime state visible in Gradle output and sandbox files, adding conservative conflict detection, supporting a random backend port when requested, and introducing `runIdeSplitMode` as a convenience task that starts the backend, waits for the join link, and launches the frontend with the correct frontend sandbox.

The result is a split-mode workflow that is more diagnosable for humans and more actionable for coding agents: the console now gives enough evidence to decide whether to wait, reconnect, kill a stale backend, change a port, or report a real startup failure.

## What Changed

- Added `splitModeServerPort` to `RunIdeTask`.
- The default remains `5990`, preserving the existing split-mode backend port.
- Setting `splitModeServerPort = 0` selects and logs a random free port in `5990..6989` for that execution.
- `runIdeBackend` now passes the resolved port explicitly as `serverMode -p <port>`.
- The resolved port is cached for the task execution so repeated reads use the same value.
- The backend join link file continues to be written from backend output, including selected-port behavior covered by tests.

- Added sandbox PID file providers.
- `split-mode-backend.pid` records the backend process while it is running.
- `split-mode-frontend.pid` is available for symmetry and future observability, although this PR only writes the backend PID.

- Added backend PID tracking for `runIdeBackend`.
- The task starts a small daemon tracker while Gradle runs the IDE process.
- The tracker looks for a newly appeared descendant process running the split-mode backend command.
- When found, it writes the PID to `split-mode-backend.pid` and logs a lifecycle message with the PID, port, and PID file path.
- The PID file is removed when the task exits so it does not intentionally point at a dead backend.

- Added conservative conflict detection before launching a new backend.
- If `split-mode-backend.pid` points to a live IDE-like process, `runIdeBackend` fails fast with the PID, optional command, PID file path, and suggested recovery.
- If the PID file is stale, malformed, dead, or cannot be positively verified as an IDE process, it is treated as stale and removed.
- If the configured port appears to be in use, the task now warns instead of blocking the launch.
- On macOS/Linux the warning includes best-effort listener PIDs from `lsof`; on platforms where that is unavailable, it still reports that the owner could not be determined.

- Added lifecycle-level sandbox path logging.
- `runIdeBackend`, `runIdeFrontend`, and the combined split-mode flow now print the effective `idea.config.path`, `idea.system.path`, `idea.log.path`, and `idea.plugins.path` values.
- This makes sandbox selection visible directly in Gradle output, which helps diagnose path mixups and makes agent-run logs self-contained.

- Added `runIdeSplitMode`.
- The task starts the split-mode backend and then launches the JetBrains Client frontend once the backend join link is available.
- The backend and frontend are launched through Gradle-managed `JavaExecSpec` instances rather than direct `ProcessBuilder` usage.
- The task accepts the expected split-mode shutdown exit code `193` for both processes.
- The frontend launch uses the `runIdeFrontend` sandbox directories and a generated frontend properties file, avoiding backend/frontend sandbox reuse.

- Fixed combined split-mode debug behavior.
- The backend launched by `runIdeSplitMode` still receives the task JVM debug configuration.
- The frontend launched by the same combined task filters copied JDWP arguments and disables copied Java debug options so it does not try to reuse the backend debugger connection.
- Standalone `runIdeFrontend` remains debuggable.
- Backend debug join links such as `debug://localhost:<port>#remoteId=Split%20Mode` are preserved and remain separate from frontend JVM JDWP debugging.

- Added and expanded tests for the split-mode workflow.
- Default port is passed to `serverMode`.
- `splitModeServerPort = 0` selects a random free port and logs it.
- Backend PID logging is observable.
- A live recorded backend PID fails fast.
- An occupied configured port warns but still proceeds.
- A stale backend PID file is ignored.
- `runIdeSplitMode` starts backend before frontend.
- `runIdeSplitMode` reuses the configuration cache.
- Expected shutdown exit code `193` is accepted.
- Unexpected frontend exit codes fail the task.
- Combined frontend launch uses frontend sandbox paths and properties.
- Combined debug launch keeps JDWP only on the backend process.

## Why

Split-mode backend startup can fail or hang in ways that are hard to interpret from the Gradle side, especially when used by automation or coding agents. The important state lives outside Gradle's normal task model: an IDE process, a server port, a join link, sandbox paths, and sometimes an orphaned backend left from an earlier run.

This PR makes that state explicit. Clear messages, PIDs, PID files, ports, sandbox paths, and a one-command combined startup task give agents and users enough information to decide whether to wait, kill a stale backend, choose another port, reconnect the frontend, or report a real startup failure.

## Backward Compatibility Analysis

The intended compatibility posture is conservative: normal launches should continue to launch as before. This PR is meant to add observability, convenience, and fail-fast behavior only for problems that were previously hidden, not to alter successful startup behavior.

What stays compatible:

- The default split-mode backend port is still `5990`.
- Existing builds that do not configure `splitModeServerPort` continue to use the same port.
- Existing `runIde`, `runIdeBackend`, and `runIdeFrontend` tasks remain available.
- Standalone `runIdeFrontend` keeps its Java debugging behavior.
- Backend debug join-link behavior is preserved.
- `runIdeBackend` is still executed through Gradle's managed `JavaExec`; the PR does not replace it with `ProcessBuilder` or a custom launcher.
- The combined `runIdeSplitMode` task also uses Gradle-managed exec operations for both backend and frontend launches.
- Direct task arguments remain unsupported for split-mode backend/frontend tasks, as before.
- A port occupied by an unrelated or foreign process does not cause a Gradle-side hard failure; the task warns and lets the IDE attempt to bind, which is closer to the previous behavior before port conflict detection existed.

What changes intentionally:

- `runIdeBackend` now logs the launched backend PID when it can be detected.
- `runIdeBackend` now writes `split-mode-backend.pid` while the backend is running and removes it when the task exits.
- Starting `runIdeBackend` while a previously recorded backend from this plugin is still alive now fails fast instead of silently starting another backend that may conflict later.
- Setting `splitModeServerPort = 0` is a new opt-in behavior that chooses a free port at execution time.
- If the configured backend port is already accepting connections, Gradle prints a warning with best-effort owner information.
- `runIdeSplitMode` is a new convenience task that orchestrates backend and frontend startup; it does not replace standalone backend/frontend tasks.
- In the combined `runIdeSplitMode` task only, frontend JVM JDWP arguments copied from the backend task configuration are removed to avoid debugger connection reuse failures.

Compatibility tradeoffs:

- We keep a hard failure only for the unambiguous case: this plugin previously recorded a backend PID and that process is still alive and looks like an IDE process. That is the case where launching another backend is most likely a user-visible conflict.
- We do not hard-fail on the generic port probe. A port can be owned by an unrelated process, can be transient, or can be hard to attribute reliably across platforms. Warning preserves compatibility while still making the problem visible.
- PID reuse is handled conservatively. A recorded PID only blocks startup if the OS exposes a command line that looks like an IDE process. If the command line is unavailable, the marker is treated as stale rather than risking a false positive.
- PID tracking is best-effort. If the OS does not expose enough process information, startup still proceeds; the only loss is observability for that run.
- Listener PID lookup uses `lsof` only where appropriate and is best-effort. Failure to resolve the owner does not fail the task.
- The random port option is opt-in via `0`, so the default behavior remains deterministic for existing users.
- Frontend JDWP suppression is intentionally scoped to the frontend process launched by the combined `runIdeSplitMode` task. It does not remove backend debugging and does not change standalone `runIdeFrontend` debugging.

## Verification

- `./gradlew compileKotlin compileTestKotlin`
- `./gradlew test --tests "org.jetbrains.intellij.platform.gradle.tasks.RunIdeTaskTest"`
- `./gradlew test --tests "org.jetbrains.intellij.platform.gradle.tasks.GenerateSplitModeRunConfigurationsTaskTest"`
- `./gradlew test --tests 'org.jetbrains.intellij.platform.gradle.tasks.RunIdeTaskTest.runIdeSplitMode*'`
- Manual IntelliJ/Gradle debug launch of `runIdeSplitMode`: backend started, frontend connected, frontend could be closed and restarted, and closing all split-mode windows ended the task successfully.


kirillk:
Examples from a local Kilo split-mode run:

```text
> Task :runIdeBackend
[org.jetbrains.intellij.platform] Started split-mode backend (PID 38703) on port 60679. PID file: '<repo>/packages/kilo-jetbrains/.intellijPlatform/sandbox/kilo.jetbrains/IU-2026.1/split-mode-backend.pid'.
[org.jetbrains.intellij.platform] Split-mode backend sandbox paths:
  idea.config.path=<repo>/packages/kilo-jetbrains/.intellijPlatform/sandbox/kilo.jetbrains/IU-2026.1/config_runIdeBackend
  idea.system.path=<repo>/packages/kilo-jetbrains/.intellijPlatform/sandbox/kilo.jetbrains/IU-2026.1/system_runIdeBackend
  idea.log.path=<repo>/packages/kilo-jetbrains/.intellijPlatform/sandbox/kilo.jetbrains/IU-2026.1/log_runIdeBackend
  idea.plugins.path=<repo>/packages/kilo-jetbrains/.intellijPlatform/sandbox/kilo.jetbrains/IU-2026.1/plugins_runIdeBackend
Join link: tcp://127.0.0.1:60679#...
```

```text
> Task :runIdeFrontend
[org.jetbrains.intellij.platform] Split-mode frontend sandbox paths:
  idea.config.path=<repo>/packages/kilo-jetbrains/.intellijPlatform/sandbox/kilo.jetbrains/IU-2026.1/config_runIdeFrontend/frontend
  idea.system.path=<repo>/packages/kilo-jetbrains/.intellijPlatform/sandbox/kilo.jetbrains/IU-2026.1/system_runIdeFrontend/frontend
  idea.log.path=<repo>/packages/kilo-jetbrains/.intellijPlatform/sandbox/kilo.jetbrains/IU-2026.1/log_runIdeFrontend/frontend
  idea.plugins.path=<repo>/packages/kilo-jetbrains/.intellijPlatform/sandbox/kilo.jetbrains/IU-2026.1/plugins_runIdeFrontend
```

